PR #23, opened 2017-02-06
MIA PT Orestes-Garcia

ogm710811:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,76 @@
...
+  font-size: 20px;
+  font-weight: bold;
+  line-height: 150%;
+  padding: 20px 100px;
+}
+
+#recipe-info {
+  max-width: 100%;
+  height: 80px;
+  padding-left: 100px;
```
nickborbe: Not sure why you have the padding here.  This is making it off center.  If you just take it off, it will be centered better.

nickborbe:
Overall, it looks good.  A couple issues with keeping things centered/in the right position.  Soon, we will introduce some tools that will help make it easier to position things in the right place.

PR #24, opened 2017-02-06
Sholpan-Kruger-done-draft1

sholpank:
My top background image isn't really fitting well need margin on the right. I tried to add but it isn't adding for some reason.

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,101 @@
+body {
+font-family: Helvetica, Arial, sans-serif;
+background-color: #fff;
+color:#333;
+font-weight: lighter;
+}
+
+#main-container {
+  margin: 0 auto;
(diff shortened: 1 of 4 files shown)


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,105 @@
...
+background-color: #fff;
+color:#333;
+font-weight: lighter;
+font-size: 18px
+}
+
+#main-container {
+  margin: 0 auto;
+  width: 100%;
+  margin-left: 30px;
```
nickborbe: This is why the whole page is off center.  No need to add a margin on the left.

nickborbe:
Other than the whole page being a little of center, everything else looks very good.

starter-code/apple-pie-starter-code/css/main.css
sholpank: cool, thanks!

PR #30, opened 2017-02-07
MIA PT Sherwin Garcia

sherwino:
90% Done


sherwino:
I wanted to make one more change, and remove the IMG tag altogether.

PR #31, opened 2017-02-07
MIA PT Colby Allen

colbyallenc:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
...
+margin: 0px;
+  font-family: 'Open Sans', sans-serif;
+  font-size: 18px
+}
+
+figure {
+margin: 10px;
+}
+
+.applePies {
```
nickborbe: This is one of the things that is causing your webpage to be really wide and have a horizontal scroll.  The reason for this is that this element is a block element, and when you move it over on the page, it continues taking up all the width a block element normally takes up, but it is moved over so it continues taking up space off the page to the right.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
...
+  position: relative;
+  left: 400px;
+  top: -550px;
+  color: white;
+  font-size: 90px;
+
+}
+
+.main {
```
nickborbe: This is also making your webpage super wide.  This thing has a default width, and if you move it over to the right like you did, you have to adjust the width and make it skinnier so that it fits.

PR #33, opened 2017-02-08
MIA PT WADSON

iengineer:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,62 @@
...
+/* Hero Image (Apple Pie) */
+#hero-image img {
+  width: 100%;
+  height: 715px;
+}
+
+.hero-text {
+  color: #fff;
+  font-size: 5em;
+  position: absolute;
```
nickborbe: Position absolute using pixels will make your web-page unresponsive.  Not the focus of this assignment but keep in mind for future reference.

nickborbe:
Looks like you got through most of it pretty smoothly.  Your one problem area seems to be positioning things in the right places.  Don't worry.  That's probably the hardest part of this.  In the coming weeks, we'll introduce CSS frameworks like Bootstrap to help you with this.

PR #34, opened 2017-02-08
Mia pt-Jessica.Palacios

MzLovelyj:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,67 @@
+body {
```
nickborbe: Good stuff overall.  If anything, the thing you have a little trouble with is putting things in the exact right place.  We will learn some new tools in the coming weeks that should help with that!

PR #36, opened 2017-02-08
Marisa-Bruno 

marisabruno:


Week 1/Day 3/recipes/css/styles.css
```diff
@@ -0,0 +1,68 @@
+.apple-recipe-header{
+    background-image: url(../images/apple-pie.jpg);
+    background-size:cover ;
```
nickborbe: Background-images. Background-size cover. You'll end up using this a million times.  Also see for future use `background-repeat: no-repeat;`

Week 1/Day 3/recipes/css/styles.css
```diff
@@ -0,0 +1,68 @@
...
+    height:650px;
```
nickborbe: I might use a percentage for the height for this so that the page can be viewed on a smaller window.  But we will go over this type of stuff in the next week or two.

Week 1/Day 3/recipes/css/styles.css
```diff
@@ -0,0 +1,68 @@
...
+
+.recipe-overview p {
+    padding:50px 200px;
+    font-size: 24px;
+    font-family:'Lato', sans-serif;
+    font-weight: 300;
+    text-align:justify;
+}
+
+.recipe-info img {
```
nickborbe: Fancy selector here.  Do you know what this does exactly?

Week 1/Day 3/recipes/css/styles.css
```diff
@@ -0,0 +1,68 @@
...
+.ingredient-list p{
+    font-family:'Lato', sans-serif;
+    font-weight: 300;
+    font-size: 18px;
+    padding-left: 225px;
+    padding-right:225px;
+    padding-bottom: 50px;
+}
+
+.cook-time img {
```
nickborbe: The same fancy selector.  You're using correctly so I guess you do know what it does.

Week 1/Day 3/recipes/index.html
```diff
@@ -0,0 +1,136 @@
+<!DOCTYPE html>
```
nickborbe: You're putting everything inside of `<div>` tags.  a lot of people don't instinctively do this when they start out.  👍

PR #40, opened 2017-02-11
MIA PT Juan Duque

imth3walrus:


nickborbe:
Looks really good.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,47 @@
+body {
+  font-family: 'Open Sans';
+  font-size: 18px;
+  color: #444;
+  margin:3% 16% 3% 16%;
+}
+
+.banner {
```
nickborbe: Background-image - good idea.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,47 @@
+body {
+  font-family: 'Open Sans';
+  font-size: 18px;
+  color: #444;
+  margin:3% 16% 3% 16%;
```
nickborbe: This is an interesting way to center al the contents on the page.  It won't work on a larger project where some things are centered and some are not, but works fine here.

PR #43, opened 2017-02-18
MIA PT Miguel

migmaldonado:
Submitting. thanks!


nickborbe:
Thanks for sending this.
Looks good mostly, a couple things less than perfect, but you're Instagram clone looks better.

PR #65, opened 2017-05-31
First upload of recipes

alidagc:


kehv1n:
Very clean code Alida. Nice div Structure and selectors. You can use more intricate selectors by reading the w3 CSS selectors doc. Great job regardless.

PR #66, opened 2017-05-31
mia robert jones

jsrj:


kehv1n:
Nice HTML structure. Nice use of DIVS. Try to use more intricate CSS selector (read w3 schools CSS selectors). We'll work on the structure inside the divs a bit too. Good job regardless RJ

PR #67, opened 2017-05-31
MIA crystal-jones

CrystalCodes01:


kehv1n:

kehv1n:
Also make sure all of your elements are inside divs or they will be killed by the div monster..

CrystalCodes01:
Do you mean put the images in divs?

PR #71, opened 2017-05-31
MIA Gabriela Parra

gbyparra:
MIA Gabriela Parra


kehv1n:
Nice job on the HTML structure. Nice use of divs. Check out some more intricate CSS selectors. Read some CSS selector docs on w3 schools. Great job none the less.

PR #75, opened 2017-05-31
MIA Jessica Cove

jessicaacove:


kehv1n:
Nice div structure Jessica. Try to use some more intricate CSS selector (read the docs). GJNTL

PR #78, opened 2017-05-31
MIA Adrian

LeyLine:


kehv1n:
I like the div structure. Keep working on making sure you split everything up the right way with the divs and using the right selectors. Gjntl

PR #80, opened 2017-05-31
MIA Nik Echenique

Nik-E:


kehv1n:
Great job Nick. I like the div structure. Check out some more CSS selectors like the child selector and selecting elements from Inside a div ( i.e <div class = 'blah'><p>Hello</p></div> //// .blah p {color: red } )

Gjntl

PR #81, opened 2017-05-31
MIA Sartaj

sajrawat:
Not finished, in progress


kehv1n:
Great job Taj, you're making good progresss on the divs. Check out some <hr> tags to make fancy lines like that. And stay away from inline styling. Make sure you close your tags where they need to be closed. Gjntl

PR #82, opened 2017-05-31
valerie-duran done

valzzie:


kehv1n:
Nice use of divs Val. Check out some more intricate CSS selectors by reading some of the docs (CSS selectors). Gjntl

PR #84, opened 2017-05-31
MIA Nicholas Hoats

hollywood017:


kehv1n:
Nice use of the main container & bodyContainer nick, great job. Try to make more use of your CSS selectors, read the selectors doc on w3. Gjntl

PR #89, opened 2017-06-12
julius-suominen

jsur:


matilore:
Good html structure and good use of css selectors, in case you have more time try to work more on the css to make it even more similar to the original one, but very good job anyway

PR #130, opened 2017-06-27
Paris Octave recipe clones

octlap:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,96 @@
...
+}
+
+#general-info p {
+  width: 60%;
+  text-align: center;
+}
+
+#general-info img {
+  display: block;
+  width: 25%;
```
posva: without using flex in the parent you could have added `margin: 0 auto` to center this image (because you added the display: block, otherwise it wouldn't work)

posva:
👍

PR #165, opened 2017-08-20
First ex

Digitalele:
@tawebbcn #BCN-ironhack


tawebbcn:
Good job! :+1: 

If you are curious about the right way to do the ordered list with ')' check this: https://css-tricks.com/almanac/properties/c/counter-increment/

PR #168, opened 2017-09-04
PAR - Stuart Gardner - Iteration1 

stuartogardner:
Stuart's french onion soup commit


ta-web-paris:
Good job.
You should put relative links ("css/main.css") instead of full link ("/Users/...")

PR #171, opened 2017-09-04
PAR-Thomas-Beguin-Iteration 1

liloow:


ta-web-paris:
Good job Thomas!

One improvement could be the vertical alignement of the header inside the image.

Maxence

PR #174, opened 2017-09-05
PAR - Hubert - First iteration

Hub2s:
I didn't pushed images


ta-web-paris:
The project is not finished yet but there are some good points 👍 

As I told you, you don't have to rename the files. 

But they will have other exercices to train!

Maxence

PR #232, opened 2017-10-17
MIA Alejandra-Del Valle

alexdelvalle:
Would love to have feedback on whether I'm using an appropriate method of positioning text/images. I feel like I struggle mostly with that in the sense that I'm not using the most efficient code for it.

Thanks!


jessicaalin:
Flex and grid is helpful for positioning ;) haha

PR #244, opened 2017-10-20
Paris - David Cohen

dmc1985:


ta-web-paris:
Good job!
You could have improved it by centering your text into your image.
Max

PR #245, opened 2017-10-21
Commited first HTML exercise - Charlotte Paris Ironhack

Charlottebtmn:
Pas tout à fait fini mais je te le donne quand même :)


ta-web-paris:
Good job and I'm sure with more time you would have solved it :)
Max

PR #246, opened 2017-10-21
Paris/Luca/apple-pie-exos

lucamerzi:


ta-web-paris:
Good jobs, congrats!
Be careful to put relative links (and not "///file/...")

PR #299, opened 2018-01-09
BCN - Jana

jjanastinn:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,63 @@
+body {
```
```
/* ---- reset ---- */
h1, body, html, main {
  margin: 0;
}
```

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,63 @@
...
+    font-family: sans-serif;
+}
+
+div {
```
tawebbcn: Avoid using element selectors (div, p, ul, h1, ol...), better use classes and ids.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,63 @@
...
+
+div img {
+    width: 50%;
+    display: block;
+    margin: 20px auto 20px auto;
+}
+
+h1 {
+    position:absolute;
+    transform: translateX(-50%);
```
tawebbcn: There are many ways to center elements, but usually you are going to use 
```
text-align: center; /* (inline elements) */
```
or 
```
margin: 0 auto; /* (block element) */ 
```
Here you have an amazing guide: https://css-tricks.com/centering-css-complete-guide/

PR #301, opened 2018-01-09
BCN - Alla Roman

allarom:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,106 @@
...
+    font-family: 'Open Sans', sans-serif;
+    font-size: 1.1em;
+    font-weight: 300;
+    line-height: 1.5em;
+    color: rgb(75, 75, 75);
+    margin: 1.1em;
+	padding: 1em;
+}
+
+p, h2 {
```
tawebbcn: Avoid using element selectors (div, p, ul, h1, ol...), better use classes and ids.

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,43 @@
...
-3 tablespoons all-purpose flour
-1/4 cup water
-1/2 cup white sugar
-1/2 cup packed brown sugar
-8 Granny Smith apples - peeled, cored and sliced
-
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
+<h2>Directions</h2>
+<p>1) Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
```
tawebbcn: Just in case you are curious, there is a way to set your own counter with CSS, nothing to worry about, anyways.

PR #302, opened 2018-01-09
BCN - Rodrigo

rodrigodaoud:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,104 @@
...
+
+#title h1{
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 72px;
+    text-align: center;
+    color: white;
+    padding-top: 20%;
+    font-weight: lighter;
+}
+div p{
```
tawebbcn: Avoid using element selectors (div, p, h1...), better use classes and ids.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,104 @@
...
+
+/* --- component --- */
+
+.ingredients ul{
+    list-style-type: none;
+}
+.ingredients li{
+    list-style-type: none;
+}
+.ingredients ul > li {
```
tawebbcn: Be careful with the specificity of your selectors.

starter-code/french-onion-soup-starter-code/css/main.css
```diff
@@ -7,3 +10,98 @@ html {
...
+}
+.ingredients ul > li:before {
+  content: "-";
+  text-indent: -5px;
+}
+.directions ul{
+  list-style-type: decimal;
+  margin-top: 5%;
+}
+.directions ul > li{
```
tawebbcn: Just in case you are curious, there is a way to set your own counter with CSS, nothing to worry about, anyways.

PR #306, opened 2018-01-09
BCN - Pedro

pedrodiogoneto:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,33 @@
...
+    background-color: white;
+    position: relative;
+}
+
+.recipes-cooking-info {
+    text-align: center;
+}
+
+.background-pie {
+    background-image: url(/Users/pedroneto/Desktop/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg);
```
tawebbcn: Use relative paths instead of absolute ones, the same way you have done it with 'cooking-info' and 'reciping-info'. Now the background only displays in your computer.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,33 @@
+body {
```
```
/* ---- reset ---- */
h1, body, html, main {
  margin: 0;
}
```

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,33 @@
...
+    margin: 0px;
+    font-family: open sans;
+}
+
+h1 {
```
tawebbcn: Avoid using element selectors (div, p, h1...), better use classes and ids.

PR #307, opened 2018-01-09
BCN - Linda

lindaluisa:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,71 @@
+div{
+    border:1px red solid;
```
tawebbcn: Remember to get rid of this before deliver.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,71 @@
+div{
```
```
/* ---- reset ---- */
h1, body, html, main {
  margin: 0;
}
```

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+img {
+    width: 100%;
+}
+
+.container #info {
+    width: 60%
+    padding-left: 200px;
+    position: centered;
```
tawebbcn: This is not working as you expect. There are many ways to center elements, but usually you are going to use 
```
text-align: center; /* (inline elements) */
```
or 
```
margin: 0 auto; /* (block element) */ 
```
Here you have an amazing guide: https://css-tricks.com/centering-css-complete-guide/

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,71 @@
+div{
+    border:1px red solid;
+    
+}
+
+body {
+    font-family: Helvetica;
+}
+
+img {
```
tawebbcn: Avoid using element selectors (img, div, p, h1...), better use classes and ids. Furthermore, in this case you don't want all the images to be 100% width. The icons are huge!

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.container #info {
+    width: 60%
+    padding-left: 200px;
+    position: centered;
+}
+
+#cookInfo {
+ width: 60%;
```
tawebbcn: Try to be consistent with the indentation.

PR #314, opened 2018-01-09
BCN - Victor

victorpellerin:


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,13 +1,22 @@
...
 <html lang="en">
 <head>
+  <meta charset="utf-8">
+  <link rel="stylesheet" href="css/main.css">
   <title>Apple Pie Recipe</title>
 </head>
+
+<img src="images/apple-pie.jpg" alt="">
+
 <body>
```
tawebbcn: You have to use div tags!!!!!
It's all about boxes inside boxes inside boxes. We'll go in deep now. Maybe (and for sure) you'll need to use block elements like div's instead of inline elements like
h1, p, img tags to style websites

tawebbcn: You have to make sure you understand the difference between Block and Inline elements and how to style them. I leave you a little hint but feel free to google for more:

https://www.w3schools.com/html/html_blocks.asp

I also leave you an article about centering related to the concepts of inline and block elements:

https://css-tricks.com/centering-css-complete-guide/#center-horizontally-and-vertically

If you still have some doubts you can come and ask to us ! ;D

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,13 +1,22 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+  <meta charset="utf-8">
+  <link rel="stylesheet" href="css/main.css">
   <title>Apple Pie Recipe</title>
 </head>
+
+<img src="images/apple-pie.jpg" alt="">
```
tawebbcn: Maybe you need the image to be the background of your title. It could be defined as:

```
// HTML
<div class="title">
   <h1>
      Apple Pie
   </h1>
</div>

//CSS
.title{
   background-image: url("../images/your-image.png");
}
```
The background will be conditioned by the div size (because it's a background). Look for more information about backgrounds and backgroud styling in here:

https://developer.mozilla.org/en-US/docs/Web/CSS/background-image

PR #315, opened 2018-01-09
BCN - Santiago

santironhacker:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,11 @@
+body {
```
tawebbcn: Consider starting your css reseting some styles. That way you reduce browser inconsistencies. We will see more advanced stuff, for now, you can do:

```
/* ---- reset ---- */
h1, body, html, main {
  margin: 0;
}
```

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,11 @@
...
+background-color:white;
+max-width: 720px; 
+text-align: center;
```
tawebbcn: To center everything I recommend you this guide about centering in CSS

https://css-tricks.com/centering-css-complete-guide/#center-horizontally-and-vertically

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.applePie {
```
tawebbcn: Try to use id's instead of classes. Just use classes if you want to reuse the styling in mere than one field

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,16 +1,27 @@
...
 <html lang="en">
+
 <head>
+  <meta charset="UTF-8">
   <title>Apple Pie Recipe</title>
+  <link rel="stylesheet"  href="css/main.css">  
 </head>
+
 <body>
 Apple Pies
```
tawebbcn: You have to give all the elements a tag! it has to be something like:

```
<h1>
   Title1
</h1>
<p>
   paragraph blablabla
</p>
```

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,16 +1,27 @@
...
-This was my grandmother's apple pie recipe. I have never seen another one
+
```
tawebbcn: You have to use <div> tags!!!!!
It's all about boxes inside boxes inside boxes. We'll go in deep now. Maybe (and for sure) you'll need to use block elements like <div>s instead of inline elements like <h1><p> to style websites

PR #318, opened 2018-01-09
PAR - Arthur M

Qodibop:


ta-web-paris:
Thanks for submitting the exercise.

I have seen you didn't manage to display an image in CSS. The reason is that you need to add `../`. So your line 21 should look like this: `  background-image: url('../images/apple-pie.jpg');`

PR #322, opened 2018-01-23
Scott-Rowell

rowellscott:


jessicaalin:
Really great job! However, for the <ul> and <ol>, we wanted you to manipulate the symbol a little (instead of bullets use dashes, and adding a closing parenthesis to the number).

PR #323, opened 2018-01-23
Mosi Rivera - Miami

mosileko:


jessicaalin:
Great job! The only thing is that cooking-info image is stretched out a bit - that's just because you manipulated both the width and the height. If you just did one, the ratio would be preserved.

PR #324, opened 2018-01-23
[MIA][Francis - Calizo]

FrancisCalizo:


jessicaalin:
Super great job. Good on getting the image just how the example showed. Wish you would have changed the title of the pull just one more time.

FrancisCalizo:
@jessicaalin Hahaha. I definitely tried deleting the history of that with absolutely no luck

PR #325, opened 2018-01-23
Nicolas-Levo

nlevo:


jessicaalin:
Solid work! Only thing I would say is that the cover image is taking too much of the page, but nothing that playing with dimensions wouldn't fix.

PR #349, opened 2018-01-24
[MIA] - [Gyasik - Nunez]

gyasikn:
I added the overflow: hidden to the CSS. I actually had it like this before but it also does not work. It stays in the same position but does not scale to size. It just shows a zoomed in portion of each photo and they're huge.


jessicaalin:
Great job! A couple of things:
1. For Google fonts (or other similar resources), it's better to link in the head than to import in the CSS (just because @import will defer loading everything else).
2. In the French Onion, your cover image isn't displaying properly. One way to have gone about this is to play with the dimensions of the image itself since you're using the overflow property. Or you can have set the image as a background image for your cover-photo div.
Keep it up :)

PR #368, opened 2018-03-19
sebastian-nugnez

Sebachowa:


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,46 @@
...
-that it does not run off.
-
-Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
-degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
-</body>
-</html>
+    <section id="description" class="container">
+      <h3>This was my grandmother's apple pie recipe. I have never seen another one</h3>
+      <h3>quite like it. It will always be my favorite and has won me several first place</h3>
+      <h3>prizes in local competitions. I hope it becomes one of your favorites as well!</h3>
```
tawebbcn: Great looking HTML!! On my screen in specific there seems to be an issue with this area right here(I have a low resolution screen). Because they are h3, they are too big so they collapse onto each other. I also wouldn't advise to put text like this inside h3. It's still a paragraph of text so a p would be better.

PR #371, opened 2018-03-19
rapha-montenegro

raphamontenegro:
Look at the finished-work folder.


finished-work/index.html
```diff
@@ -0,0 +1,120 @@
...
+        that it does not run off.
+    </li>
+    <li>
+        Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
+        degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
+    </li> 
+  </ol>
+  </section>
+
+  <div id="recipe2">
```
tawebbcn: Try to use more descriptive and semantic html tags like section, header, footer, article, etc...

finished-work/main.css
```diff
@@ -0,0 +1,126 @@
...
+
+
+
+/* ---- components ---- */
+
+
+/* ---- had to look for the min-height thing to make the image appear as whole ---- */
+#recipe1 .recipe-picture { 
+  width:980px; 
+  min-height:500px;
```
tawebbcn: A nice trick here to make the image display as a whole without using ```height``` would be to set the ```line-height``` to the desired image height as we saw

finished-work/index.html
```diff
@@ -0,0 +1,120 @@
...
+        <div id="description" class="container">
+          <p>
+              We have been trying French onion soup in restaurants for years and my family
+              and friends agree none can compare to my recipe for taste and simplicity of
+              preparation.
+          </p>
+          <img src="/Users/raphamontenegro/Documents/ironhack/lab-css-recipes-clone/starter-code/french-onion-soup-starter-code/images/recipe-info.png" class="img-recipe-info">
+        </div>
+      </div>
+      <hr>
```
tawebbcn: Maybe a better option instead of using ```</hr>``` would be to use something like border bottom which would take the same width of the preceding div causing you to not need to style the hr after

finished-work/main.css
```diff
@@ -0,0 +1,126 @@
...
+  line-height: 147%;
+}
+
+h1 {
+  
+}
+
+h2 {}
+
+h3 {}
```
tawebbcn: If the class is empty, then you can remove it

finished-work/main.css
```diff
@@ -0,0 +1,126 @@
...
+  background-image: url("/Users/raphamontenegro/Documents/ironhack/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg");
+  padding: 30px 0;
+}
+
+#recipe2 .recipe-picture {
+  width:980px; 
+  min-height:500px; 
+  margin: 20px auto;
+  background-size: cover;
+  background-image: url("/Users/raphamontenegro/Documents/ironhack/lab-css-recipes-clone/starter-code/french-onion-soup-starter-code/images/soup.jpg");
```
tawebbcn: Here the links need to be changed too in order for it to work in every computer

finished-work/main.css
```diff
@@ -0,0 +1,126 @@
...
+
+
+
+
+/* ---- components ---- */
+
+
+/* ---- had to look for the min-height thing to make the image appear as whole ---- */
+#recipe1 .recipe-picture { 
+  width:980px;
```
tawebbcn: Avoid using width in pixels like this.

finished-work/index.html
```diff
@@ -0,0 +1,120 @@
...
+  <div class="recipe-picture">
+    <h1>Apple Pies</h1>
+  </div>
+    <div id="description" class="container">
+      <p>
+        This was my grandmother's apple pie recipe. I have never seen another one
+        quite like it. It will always be my favorite and has won me several first place
+        prizes in local competitions. I hope it becomes one of your favorites as well!
+      </p>
+      <img src="/Users/raphamontenegro/Documents/ironhack/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/recipe-info.png" class="img-recipe-info">
```
tawebbcn: Be careful when linking images and files. In this case the src is pointing to your computer. To fix this the src should look like this in your case ```../starter-code/apple-pie-starter-code/images/image-name-goes-here.png```

PR #393, opened 2018-03-20
BCN - Jordi Calzada

jordicalzada:
Almost anything done...

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,15 @@
+*{
+    margin: 0;
+    padding: 0;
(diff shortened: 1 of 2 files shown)


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,15 @@
+*{
+    margin: 0;
+    padding: 0;
+  }
```
tawebbcn: Css reset ✅ 
But don't mind on take a look on our [cheat-sheet](https://github.com/ironhack/bcn-webdev-cheatsheet/blob/master/m1/css/boilerplate.css)

PR #401, opened 2018-03-20
PAR Lucas M

lucasm-git:
Although the result's quite ok, I'm really not sure about the way I organized my css file. I obviously lack good practices or guidelines, is there a resource listing some basic rules for css organizing?
Thanks!

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,104 @@
+/***************************************************************************
+CSS RESET
+***************************************************************************/
+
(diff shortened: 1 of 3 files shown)


ta-web-paris:
Nice job, may be you can try to display the image as a cover (so it will take all the width of the window).

Regarding the css file, it looks okay, the most important thing is that you feel confident in the way you organize it, you can either organize it by element (h1, h2, img, ect..) or by blocks (nav, header, ect..)

I would recommand you to organize it by "block" so you won't struggle when you will have to change the color of a specific button per example, I'm gonna take a look and check if there is guidelines, if I find something usefull I'll let you know ! 

Keep up the good work ! 👍

PR #403, opened 2018-03-20
PAR Maggie

mbbaber:
How do you make an ordered list with "1)" instead of "1." ?


ta-web-paris:
Good job, may be you can check this documentation (https://developer.mozilla.org/en-US/docs/Web/CSS/font-size) about font and how to play with it, I'll suggest you not to use something like "250%" as an h1 font size, instead you can use "em".

Regarding the list style, you can use an UL, and remove the default list-type, and put the "1)" at the beginning of your text block.. It's not the best way but it's the only I know, anyway you won't use this kind of list a lot.

Keep up the good work, and I you have any questions, feel free to ask me !

PR #405, opened 2018-03-20
PAR Brianna

ycrtw93:
Was not sure about the font sizes, margins, and line spacing.


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
...
+
+h1 {
+  font-size: 70px;
+  text-align: center;
+  line-height: 600px;
+  letter-spacing: 5px;
+  color: white;
+  height: 600px;
+  background-size: 1000px 600px;
+  background-image: url('/Users/briannacho/code/labs/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg');
```
ta-web-paris: With this path, it only works on your computer (as you are using a path that start at the roots of your computer).

Try this instead : url('../images/apple-pie.jpg')
It will work on every computer as you are only refering to the folder path :)

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
...
+
+.title-size {
+  font-size: 25px;
+}
+
+.text-size {
+  font-size: 18px;
+}
+
+.info-pics {
```
ta-web-paris: Ok so it looks like you don't use this class in your html so you won't notice any change.

What you can do, is play with your recipe class, because it's the one that "host" almost all of your content.
And may be create a container with your recipe2 id, try with this code : 

#section2 {
  width: 80%;
  margin: 0 10%;
}

It can be a good start :)

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
...
+}
+
+h1 {
+  font-size: 70px;
+  text-align: center;
+  line-height: 600px;
+  letter-spacing: 5px;
+  color: white;
+  height: 600px;
+  background-size: 1000px 600px;
```
ta-web-paris: Try to use background-size : 100% ;)

ta-web-paris:
Good job, if you need any tips about marging, check this website, there is plenty of css tricks :) 
https://css-tricks.com/

Keep up the good work !

PR #406, opened 2018-03-20
PAR Mostafa Gad

M-Gadd:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,49 @@
+.anything {
+  background-image: url(/Users/mgad/code/labs/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg);
```
ta-web-paris: Don't use this kind of path as they will only work on your computer (because you are starting from the Roots)
Instead use the folder path like this : 

"../images/apple-pie.jpg"

It will works on every computer as it's only related to the project folder (same for your stylesheet) :)

ta-web-paris:
Be carefull with the path :)

Otherwise good job ! 👍

PR #409, opened 2018-03-20
PAR Elise

elisegriset92:
Not sure about : 
- background image and how it centers the text 
- difference of font weight in my two recepies and I don't understand why.. (mes caractères sont plus larges dans la tarte aux pommes plutôt que la soupe à l'oignon)


ta-web-paris:
I answered directly in your css file.

That's why the weight isn't exactly the same :) 
image

Good job :+1: 
Keep going!

PR #411, opened 2018-03-20
PAR Renaud

Ayss83:
In the ordered list, I couldn't display 1) 2) etc. It displayed 1. 2. etc instead. Does it only depend on the browser used?


ta-web-paris:
I think you should use :before and :after for this issue.
Source: https://css-tricks.com/almanac/selectors/a/after-and-before/

This is what I did before:
https://github.com/VivianSolide/lab-css-recipes-clone/tree/master/starter-code/apple-pie-starter-code

Good job :+1:

PR #413, opened 2018-03-20
PAR Luke

lukegeno:
Couldn't import the Open Sans font.
Margin on the top above h1/header.
Cant adjust image size (cooking-info and recipe-info) without overlapping <hr>.


ta-web-paris:
Google Font:
Take a look on it >>> https://fonts.google.com/specimen/Open+Sans?selection.family=Open+Sans
I can help you if you want (Vivian)

If you want to adjust image, you can use height: 100vh and/or 100vw
vh stands for Viewport Height
vw stand for Viewport Width

Really useful :money_mouth_face:

PR #418, opened 2018-03-20
PAR Louis 

LouisMasurel:
Hello,

- the title isn't on the image and not white
- the numbers of the last section aren't at the good place
- the position and the deformation of the text isn't right


ta-web-paris:
Hi @LouisMasurel 

1. You should use `background-image` into the first `<div>`
2. Get a look on it
```css
  ul li:before {
    content: '-';
    position: absolute;
    margin-left: -20px;
  }
```
3. I think you didn't git push correctly, I can't the issue

You can check my correction if you have time: 
https://github.com/VivianSolide/lab-css-recipes-clone/tree/master/starter-code/apple-pie-starter-code

PR #419, opened 2018-03-20
PAR Billy

Yaats:
Not sure about the way I use to display my background-image

I think I should use list for "ingredients/directions" and list-style-type in the CSS instead of writing "-" and "1)" on every line. 

Is it the same think to put my font setup (css) in a body tag instead of the html one ? 

Also I think I didn't use the right way to display my cooking-info image, cause it's not in the section tag like the recipe-info image.

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,54 @@
+html {
+  font-family: 'Open Sans', sans-serif;
+  font-weight: lighter;
+}
+
+header {
+  background-image: url("../images/apple-pie.jpg");
+  background-repeat: no-repeat;
+  background-size: cover;
+  height: 800px;
+  background-position-x: 0;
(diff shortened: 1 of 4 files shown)


ta-web-paris:
Good job! :+1: 

1. Background-image
This is my 2 cents on the issue.
vh stands for viewport height.
Pretty useful.
_vw for viewport width_
```css
.header {
    background-image: url(../images/apple-pie.jpg); 
    height: 100vh;
    width: 100%;
    vertical-align: middle;
    display: flex;
    background-attachment: fixed;
    background-position: center center;
    background-size: cover; 
}
```
2. Ul li
It's possible to change the list by a dash using this snippet :
```css
  ul li:before {
    content: '-';
    position: absolute;
    margin-left: -20px;
  }
```
3. Font
Take the wat you want.
@import into css file
link style into html

4. Image
You can easily wrap your image into <div>

PR #424, opened 2018-04-01
mia-pt-tommy-dilillo

tommydilillo:


ManuelCabrera01:
i suggest  use "<p>"tags  instead of <h2> and <h3> when you want to add plain text, is easy to manipulate and it doesn't come with many default settings

ManuelCabrera01:
i suggest to use < p >tags instead of < h 2 > and < h 3> tags when you you just want to add plain text. is easy to manipulate and it doesn't come whit many  default setting .

PR #425, opened 2018-04-01
MIA-PT-Mayk Dougllas

mdougllas:


colbyallenc:
Overall, nice job! Image is a bit disproportioned on the screen. Font looks a bit different as well. Margin/Padding is a bit smaller than the original.

PR #429, opened 2018-04-02
MIA - Daniel Lozano 

lozdan:


MarcosPalaciosIII:
good work on the apple pie and french onion clones. Could have made the banner text a little better, but overall good job.

PR #430, opened 2018-04-02
MIA-DALI-CSSDAY1

rojasdali:


ianizaguirre:
Hi, this is some feedback....

1) I recommend not using inline styles in your index.html... you did this a few times, here is an example of one:
`  `
* Instead you could have made a div and attached the image via css , for example you could have made a div called "hero" and targeted it with something like this in CSS:
```css
.hero{
  background-image: url('../images/soup.jpg');
  background-repeat: no-repeat;
  background-size: cover;
  height: 500px;
  width: 100%;
  color: white;
  margin: 0 auto;

}
```

* I saw that the entrance image you created was way bigger then the example they give you to mimic. So it was not an exact match.

2) Work on proper alignment and spacing when writing your CSS... For example, this:
```css
p{
   
    margin: auto;
    width: 60%;
    padding: 30px 0px 30px 0px;
    font-size: 22px;
    font-family: Arial;
}
```
Should be:
```css
p {  
    margin: auto;
    width: 60%;
    padding: 30px 0px 30px 0px;
    font-size: 22px;
    font-family: Arial;
}
```
* Also, I think instead of writing ` margin: auto;` you meant to write ` margin: 0 auto;` ?

3) Instead of writing a class for `hr` in HTML : `<hr class="style1">`
and then changing it via CSS:
```css
hr.style1{
    border-top: 1px solid #8c8b8b;
    width: 60%;
}
```
You could have just had no class : `<hr />`
and then changed them all with CSS:
```css
hr {
    border-top: 1px solid #8c8b8b;
    width: 60%;
}
```
* The goal with doing this is trying to remove giving so many div names to things we can target in a better way .... this way the code can look cleaner and can be read easier. 

* As a random comment , I saw You used some advanced css rules, so thats interesting :-) 
Good Job 👍 

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #434, opened 2018-04-02
FTWD Miami Osse Laborde

olaborde:


MarcosPalaciosIII:
Apple Pie:  You created a list for the ingredients but not the directions. The difference would be the type of list that you would use. Instead of an unordered list you would use an ordered list, or to make it look the same as the site your cloning you can do multiple ` <p> ` tags with `1)  2)  3) `as the beginning of each direction.  Also some padding would be required to push the main content in some more  and adjusting the size of the image showing the prep time and what not. 
* For the apple pie picture you would have to center the image before adjusting the size and change the font of the banner text. 


French Onion: 
* You seem to have deleted or removed your soup image
Again the ingredients and the directions you do not have then set up as a list. You can always do an unordered list and then do a list decoration to remove the dots it adds. afterwards add either - or #) to make it seem like the site your cloning.  Could have also added a little more padding on the main content of this one. 

Overall not bad.

PR #444, opened 2018-04-03
Nicholas - Kreps

sperkn:


colbyallenc:
Overall nice job! Font size for the title is different. The image seems to be cropped, and margins seems to be a little wider than the original.

PR #448, opened 2018-04-03
pt-mia-web-dev-mariah-dunn

teepott99:


ta-web-mia:
Hey! I see that you are having some trouble getting the images to appear- thats because of the path of the image in the css. Also, sizing of the content seems to be a little wider. 

Nice job! When you have the time, definitely go back and make the changes that you need- I can help you with any of the issues that you may encounter.

PR #453, opened 2018-04-03
MIA - Luis - css apple pie french onion clone

chiko15:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
+
+.hero {
+    background: url(../images/apple-pie.jpg);
+    padding: 25%;
+    background-size: cover;
```
sandrabosk: We want our background image to occupy only part, not the whole viewport, so we should set the height, get rid of the padding and make sure our background image has property repeat set on no-repeat. Something like this:
```
 background: url("../images/apple-pie.jpg") no-repeat center center;
height: 500px;
background-size: 100%;
```

sandrabosk:
Good job! Thank you for submitting.

PR #456, opened 2018-04-03
not done yet

rawanalfaiz:


MarcosPalaciosIII:
The banner text and image could have been sized a little more to match the site to clone. Remember you can play with the width and height for images and font-size for text. 

Overall good work on the clone

PR #458, opened 2018-04-03
mia-jens heilemann

jkai7:


ianizaguirre:
Hi, here is some feedback ...

1. You did not give the correct font sizes to your P or Heading tags. You just never added CSS `font-size: #px` to anything, so none of the font-sizes changed.

2. When you wrote , this:
```css
  .apple-pie h1{
      background-image: url("../images/apple-pie.jpg");
      background-size: 100%;
      padding: 23%;
    }
```
You did something incorrect there. You added padding so the "heading" could be in the center of the page. This is a common mistake... Padding should NOT be used to move things, the correct property to move things would be using a Margin ... so maybe a `margin-top:` or `margin-bottom:` could have been used. 

3. You never finished the French Onion exercise, but don't worry, since it was basically the same thing as the Apple pie one we did.

If you have any questions on my feedback, just send me a message on Slack. 
Aside from that, good job and keep up the good work! 😁

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #459, opened 2018-04-03
MIA-CrystalMeneses

shecancode:


ianizaguirre:
Hi Crystal, here is some feedback ...

1. Instead of writing: `  recipe-info`
You should write: `  recipe-info`
.... notice I added a `./` to the front of the URL

2. Other then that everything looks good. The only other thing I could recommend is to get into the habit of adding a "slash" to self closing tags... So for example , `<br>` is a self closing tag since you only need one tag and not a pair .... So a better way to write `<br>`  would be to write this: `<br />`.


If you have any questions on my feedback, just send me a message on Slack. 
Aside from that, good job and keep up the good work! 😁

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #460, opened 2018-04-03
MIA-Ruben Vargas

Ambition101:


MarcosPalaciosIII:
Forgot to add the pictures of the prep time and cook time to both clones. Also could have used a bit of padding on both clones as well. 
For the apple pie clone the font-size of the directions part can be modified if you want to change how big the text is. 
For the banner text, you can play around with font-size and padding if you want to put it dead center. 

overall good work.

PR #462, opened 2018-04-04
MIA-PT-18-Yandy-Gonzalez

ydnay:


colbyallenc:
I notice that you may have not had a chance to complete this. Thanks for doing the pull anyways. Whenever you get some time, definitely come back to this exercise!  
To link your images, instead of doing something like this: 
background-image: url("/Users/yandy/ironHack/labs/recipesClone/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg"); 
I would do something like this instead: 

PR #466, opened 2018-04-06
MIA-PTWD-WilliamDonayre-labCssRecipeClone

williamdonayre:
Was emailed earlier this week to Manuel


colbyallenc:
Overall nice job! I would use the following to get the margins in the correct place: 

h1, body, html, main {
  margin: 0;
}

Then give the container parts some padding.

PR #467, opened 2018-04-07
Ironhack MIA - Maximilianus P. - Part-time Web.Dev.

XamSP:
Finally was able to request a pull again!


colbyallenc:
You are having some trouble linking the images- there is an error in your path. I would try something like this:
That should work.

PR #517, opened 2018-05-30
PAR Jihad

eloji9:
What I wanted to do was to go step by step to make one of the recipes then apply the same principles to the second one.
I was not able to finish as always, which is frustrating.
The elements I noted for now are difficulties about cropping images and resizing them. I should probably look for other solutions.

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,49 @@
+header{
+    border-style: solid;
+}
+body {
+    border-style: solid;
+    margin: 200px;
+}
+.presentation1 {
+    border-style: solid;
+    position: relative;
+}
+.h1-centered{
+    border-style: solid;
+    position: absolute;
+    top: 30%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    color: white;
+    font-size: 200%;
+    font-family: 'Gill Sans', 'Gill Sans MT', Calibri, 'Trebuchet MS', sans-serif;
+}
+
+.presentation2 {
+    border-style: solid;
+    padding: 10%;
+    text-align: justify;
+    padding: -50xp;
+    font-size: 120%;
+    font-family: Arial, Helvetica, sans-serif;
+}
+.image1{
+    border-style: solid;
+    max-width: 100%; overflow: hidden;
+    text-indent: -100px -300px -100px -100px;
+    }
+.image-recipe{
+    border-style: solid;
+    padding: 0px 150px;
+    width: 100px;
+}
+hr {
+    width = 75%;
+    margin-left: auto;
+    margin-right: auto;
+    border-style: solid;
+    border-width: 1px;
+    border-color: grey
+
+}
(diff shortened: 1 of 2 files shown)


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,49 @@
...
+    padding: 10%;
+    text-align: justify;
+    padding: -50xp;
+    font-size: 120%;
+    font-family: Arial, Helvetica, sans-serif;
+}
+.image1{
+    border-style: solid;
+    max-width: 100%; overflow: hidden;
+    text-indent: -100px -300px -100px -100px;
```
ta-web-paris: try not to use negtive value as it will be a mess if you are building a responsive website, instead try to use the position & the flexbox method that we've seen yesterday

starter-code/apple-pie-starter-code/index.html
```diff
@@ -2,34 +2,51 @@
...
-1/2 cup white sugar
-1/2 cup packed brown sugar
-8 Granny Smith apples - peeled, cored and sliced
+  <hr width= 300px>
 
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
-flour to form a paste. Add water, white sugar and brown sugar, and bring to a boil. Reduce
-temperature and let simmer.
+  <div class = "ingredients"><h2>Ingredients</h2>
```
ta-web-paris: You don't have to wrap your h2 in a div, if you want to add specific styling to this element just add a class to the h2, because to many may lead to some weird results (box in box in box..)

starter-code/apple-pie-starter-code/index.html
```diff
@@ -2,34 +2,51 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="stylesheet" type="text/css" media="screen" href="./css/main.css" />
+  <script src="main.js"></script>
 </head>
-<body>
-Apple Pies
-This was my grandmother's apple pie recipe. I have never seen another one
-quite like it. It will always be my favorite and has won me several first place
-prizes in local competitions. I hope it becomes one of your favorites as well!
+  <header><div class = "presentation1"><div class="image1"> <img src = "../apple-pie-starter-code/images/apple-pie.jpg" alt="Apple pie" style="width:100%;"></div>
```
ta-web-paris: for this one there is a trick, you can create a div, with the h1 inside, and add the following styling code : 

```css

#header_soup {
  background-image: url("../french-onion-soup-starter-code/images/soup.jpg");
  background-size: cover;
  height: 600px;
  margin: 0 5% 0 5%;
  background-attachment: fixed;
}
``` 

as you can see I used the background-image property (so it was way easier to add the text and to play with it) and to resize the image without struggling with cropping or stuff like this, just use the "background-size : cover" :)

ta-web-paris:
Sometimes playing with image can be really boring and frustrating, if you want to keep the ratio of an image you can use this small trick : 

```css

image {
width: 300px (per exemple);
height: auto (using auto will resize the image depending on the absolute value, it can be the height or the width)
}

```

Otherwise good job, don't use to many div as mentionned before, and before rushing in the code try to regroup block element (as you can see the part with the ingredient, and the directions part have the same layout, so you can start by wrapping them in a big div and then place the small element) :)

Let me know if you have any question, I'll be happy to help

PR #518, opened 2018-05-30
PAR Ghaith & Meri

mpligioso:
@GhaithHan


ta-web-paris:
Your code is good, I don't have any comment for the moment, you manage to structure your content in a good way, you are not using too many CSS rules (which is nice :) )

Good job! 💯 If you have any question regarding css / html and what we've seen so far let me know :)

PR #519, opened 2018-05-30
PAR Shanshan Cao

moonchie:
The two CSS files are exactly the same, except for the banner background picture src.

### DIFF
--- a/starter-code/apple-pie-starter-code/index.html
+++ b/starter-code/apple-pie-starter-code/index.html
@@ -1,35 +1,69 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <title>Apple Pie Recipe</title>
+  <title>Apple Pie recipe</title>
+  <!-- Google fonts gives us access to the Open Sans Font -->
+  <!-- Google fonts is also awesome! Make sure to check it out -->
+  <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,600" rel="stylesheet">
+  <link href="css/main.css" rel="stylesheet">
+
 </head>
+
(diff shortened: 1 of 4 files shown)


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,69 @@
...
-
-Place the bottom crust in your pan. Fill with apples, mounded slightly. Cover with a
-lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
-that it does not run off.
-
-Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
-degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
+    <!-- Banner-->
+      <div class = "banner-container">
+        <div class = "soup">
```
ta-web-paris: Only one div is enough to wrap you h1 & add the background image, and as you are not referring to the div class soup in your css you can remove it it will not affect your design :)

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,69 @@
...
+          <h1> Apple Pies</h1>
+        </div>
+      </div>
+
+<!-- Recipe-->
+<div class = "recipe-container">
+    <div class = "recipe-content">
+      <h3> This was my grandmother's apple pie recipe. I have never seen another one
```
ta-web-paris: That's the only point I see, you should use a <p> tag instead of the <h3> here, as the h tag are helpful when it comes to organize and add some hierarchy to your content

PR #520, opened 2018-05-30
PAR Marie & Diane

Dianedel:
@marieglr

### DIFF
--- a/starter-code/apple-pie-starter-code/index.html
+++ b/starter-code/apple-pie-starter-code/index.html
@@ -1,35 +1,52 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
+  <meta charset="UTF-8">
+  <link rel="stylesheet" href="./css/main.css">
   <title>Apple Pie Recipe</title>
 </head>
+
 <body>
-Apple Pies
-This was my grandmother's apple pie recipe. I have never seen another one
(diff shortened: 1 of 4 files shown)


Dianedel:
We will finish later.

Dianedel:
We have finish our work with this second version. Thank you

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,52 @@
...
+      <li>1/4 cup water</li>
+      <li>3 tablespoons all-purpose flour</li>
+      <li>1/4 cup water</li>
+      <li>1/2 cup white sugar</li>
+      <li>1/2 cup packed brown sugar</li>
+      <li>8 Granny Smith apples - peeled, cored and sliced</li>
+    </ul>
+    <img class="image" src=".\images\cooking-info.png" alt="cooking info">
+  </section>
+  <hr width="75%">
```
ta-web-paris: that's the only point where I can say something, you should keep the styling in the css (even if it's work like this) because on a big project you may loose yourself if you add piece of styling everywhere, better to have all in the same place :)

ta-web-paris:
Your code is well structured, like the use of the section tag :)

Good job! Continue like this 💯

PR #535, opened 2018-06-11
done

se66e:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,60 @@
...
+}
+
+hr {
+margin-left: 20px;
+margin-right: 20px;
+}
+   
+
+.recipe-header {
+  background-image: url("/Users/sebastianbjers/2018-06/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg");
```
tawebbcn: Don't write the complete path.
We can't access what's inside your computer. 
Just go for "./images/apple-pie.jpg"

PR #591, opened 2018-06-12
MIA christine beck

cpbcpb:
Christine Beck Lab 1 recipe clone
Still need to figure out how to change ordered list and numbered list style to - and 1).


ta-web-mia:
to modify the bullets points in a 'ul' or 'ol' element you can apply a rule call {list-style-type: none}
that will remove ani style including dots and number after that you can add any style that you want.

PR #639, opened 2018-07-30
BER - Tormod - done

Ojself:


ta-web-ber:
Very good job for the first day

Silvio

PR #648, opened 2018-07-31
BER - Jason - Done

JasonSimms:


ta-web-ber:
Hey Jason! First commit! It's a job well done, only  a few minor notes-but again these aren't huge, because I think you got the concepts well here. 

- Image URL's in HTML start from the folder that the HTML file is in. And don't start with a slash here :)
   VS 

- you made the entire body 50% wide and centered it, which is fine in itself because it's extra. But you forgot to do this with the recipe images!

Small stuff really! Good job.

ta-web-ber:
I see i ws not allowed to include the code for the img tags so,
src="images/recipe-info.png" VS src="starter-code/apple-pie-starter-code/images/cooking-info.png"

should get the same point across. If not, just ask

PR #656, opened 2018-08-07
PAR Pauline & Constance

Cr27910:
@paulettepaillette


ta-web-paris:
Great job! Good call to use css "border-bottom" property rather than an <hr> in the html, and you manage to have an unordered list with "-", bien joué! :100:

PR #659, opened 2018-08-07
PAR Mohamed & Abi

SMoZam:
@abisummers

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,87 @@
+*{
+
+margin: 0;
+padding: 0;
+
+}
+
+body{
+line-height: 1.6;
+font-family: ‘Montserrat’, sans-serif;
+
+}
+
+
+h2 {
+    padding-bottom: 20px;
+}
+
+.cover {
+    background-image: url(../images/apple-pie.jpg);
+    background-size: cover;
+    background-repeat: no-repeat;
+    height: 1100px;
+    width: 100%;
+    margin-bottom: 50px;
+    text-align: center;
+    color: white;
+
+}
+
+.cover h1{
+
+position: absolute;
+top: 50%;
+left: 40%;
+font-size: 5em;
+
+}
+
+
+.applepie-background img{
+width: 100%;
+height: 100%;
+}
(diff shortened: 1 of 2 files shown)


starter-code/apple-pie-starter-code/index.html
```diff
@@ -2,34 +2,66 @@
...
+
+
+
+  <section class="description">
+  
+    <p>  This was my grandmother's apple pie recipe. I have never seen another one
+    quite like it. It will always be my favorite and has won me several first place
+    prizes in local competitions. I hope it becomes one of your favorites as well! </p>
+    <img src="images/recipe-info.png">
+    <hr>
```
ta-web-paris: most html tags have a semantic meaning, the meaning of the hr tag is thematic break, it is a bit redundant with the use of different sections. It is better to keep in the css everything that has only style purpose: you can recreate this horizontal line in the css by adding a border-bottom to the section

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,87 @@
...
+    text-align: center;
+}
+.ingredients{
+    
+    text-align: left;
+    margin: 0 10%;   
+}
+
+.ingredients li {
+    list-style: '-' ;
```
ta-web-paris: Good call to use a ul with lis for this part, and to try to change the list style type in the css!
since '-' is not a css built-in list style type, the syntaxe to do that however is a bit more complicated and it won't work by just hardcoding it like that.  This is the correct syntax to do that, but you will learn about that later in the course : 
.ingredients li::before{
    content:"- ";
  }

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,87 @@
...
+
+.ingredients img{
+
+display: block;
+margin: 0 auto;
+
+}
+
+
+.ingredients-order{}
```
ta-web-paris: to keep your code clean, you can erase or comment out tags that you ended up not using :)

ta-web-paris:
Overall good job! Check the few comments, and Flexbox is going to be super useful for managing margins and positions :)

PR #664, opened 2018-08-07
PAR Julie & Aurélien

LOOFFI:
@youlie03

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,43 @@
+
+* {
+  font-family: 'Open Sans', sans-serif; 
+  font-size: 30px;
+}
+
+.header {
+  background-image: url(file:///home/looffi/code/labs/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg);
+  width: 100%;
+  background-size: cover;
+  height: 700px;
+}
+
+h1 {
+  font-size: 80px;
+  color: white;
+  font-weight: 500;
+  text-align: center;
+  padding-top: 350px; 
+}
+
+hr {
+  margin: 40px 300px;
+  height: 6px;
+  background-color: gray;
+}
+
+#intro p {
+  padding: 0 100px;
+}
+
+#Ingredients{
+  padding: 0 100px;
+}
+#directions{
+  padding: 0 100px;
+}
+ul{
+  list-style-type: none;
+}
+#directions p{
+  margin: 0 0;
+}
--- a/starter-code/apple-pie-starter-code/index.html
+++ b/starter-code/apple-pie-starter-code/index.html
@@ -1,35 +1,58 @@
 <!DOCTYPE html>
 <html lang="en">
+    <link href="https://fonts.googleapis.com/css?family=Open+Sans" rel="stylesheet"> 
+<link rel="stylesheet" href="./css/main.css">
 <head>
   <title>Apple Pie Recipe</title>
 </head>
 <body>
-Apple Pies
-This was my grandmother's apple pie recipe. I have never seen another one
-quite like it. It will always be my favorite and has won me several first place
-prizes in local competitions. I hope it becomes one of your favorites as well!
+  <header class="header">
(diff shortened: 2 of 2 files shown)


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,43 @@
+
+* {
+  font-family: 'Open Sans', sans-serif; 
+  font-size: 30px;
+}
+
+.header {
+  background-image: url(file:///home/looffi/code/labs/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg);
```
ta-web-paris: Using this path for the image will work on your computer, but not on others (not on mine for instance), and not when you push your projects online!
This is why your image paths should be "relative", meaning the path goes from your index.html file to your image file inside your folders. Here, the correct path would be: `'../images/apple-pie.jpg'`

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,43 @@
...
+  background-size: cover;
+  height: 700px;
+}
+
+h1 {
+  font-size: 80px;
+  color: white;
+  font-weight: 500;
+  text-align: center;
+  padding-top: 350px;
```
ta-web-paris: Here, adding some `padding-top` is not a very good solution because because it will not scale with different image sizes. There's one very precise way of doing it: `translate`.

Here's what your code could look like:
```css
h1 {
  font-size: 80px;
  color: white;
  font-weight: 500;
  text-align: center;
  position: relative;
  top: 50%;
  transform: translateY( -50% );
}
```

This technique also works very well for other element centering inside containers

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,58 @@
...
 <head>
   <title>Apple Pie Recipe</title>
 </head>
 <body>
-Apple Pies
-This was my grandmother's apple pie recipe. I have never seen another one
-quite like it. It will always be my favorite and has won me several first place
-prizes in local competitions. I hope it becomes one of your favorites as well!
+  <header class="header">
+    <h1>Apple Pies</h1>
```
ta-web-paris: Since your container already is a header, you don't really need to give it a class of "header". You can simply target the entire container like so:
```css
header {
    /* your styling here */
}
```

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,58 @@
...
+  <hr >
+</section>
 
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
+<section id="Ingredients">
+  <h2>Ingredients</h2>
+  <ul>
+    <li>- 1 recipe pastry for a 9 inch double crust pie</li>
+    <li>- 1/2 cup unsalted butter</li>
```
ta-web-paris: If you want to style your bullet points there's a more "official" way to do it, with the `list-style-type` property. It has quite a lot of values, you can learn more about it here: https://developer.mozilla.org/fr/docs/Web/CSS/list-style-type

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,58 @@
...
+  </ul>
+  <center><img src="./images/cooking-info.png" alt="cooking-info img"></center>
+  <hr>
+</section>
+
+
+<section id="directions">
+  <h2>Directions</h2>
+<p>1) Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
 flour to form a paste. Add water, white sugar and brown sugar, and bring to a boil. Reduce
```
ta-web-paris: Here, instead of creating separate `<p>` tags with "1)", "2)" and "3)", an ordered list would have been a better (and way more scalable) solution.
As a reminder, it works exactly as a `<ul>` but it's written `<ol>`, and instead of bullet points it creates numbers

ta-web-paris:
Your page looks good, and CSS general concepts are well understood :)

Keep up the good work!
Lucas

PR #667, opened 2018-08-13
MIA FTWD Nick Crispino

crispinonicky:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,43 @@
...
+img {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+hr{
+  width: 75%;
+}
+
```
MarcosPalaciosIII: too many repeating styles. A good work around is to wrap the tags that are together in a div and then adjust the margin and/or padding of the div

starter-code/french-onion-soup-starter-code/css/main.css
```diff
@@ -7,3 +7,42 @@ html {
...
+img {
+  display: block;
+  margin-left: auto;
+  margin-right: auto;
+}
+
+hr{
+  width: 75%;
+}
+
```
MarcosPalaciosIII: too many repeating styles. A good work around is to wrap the tags that are together in a div and then adjust the margin and/or padding of the div

MarcosPalaciosIII:
good work on the recipe clone. Just remember when adjusting multiple tags position that are in the same section, you can wrap the tags in the div tag and control the div. Then you can adjust the spacing of the tag's within the div with padding and whatever additional style the tags may need. Good practice is to give each tag a class so that when editing that class it will not conflict with styling you may wan't to give another tag of the same type (ie: 3 p tags that need different spacings).

keep up the good work.

PR #675, opened 2018-08-13
Jonathan MIA css clone apple and onion

yonathanh:


sandrabosk:
Good work in the first part (apple pie) but still missing a lot of styles on the second part of the exercise. If you have time, fix the second part. Thank you for submitting.

PR #696, opened 2018-08-20
[MAD DE] Matt-Rocha

mattchrocha:
- The exercise was not too complicated and allowed to review consistently inline and block elements.
- Although I understand it was probably not as important, it was a little frustrating to work without the correct fonts, weights, etc.
- After today's Ana's comment about flexboxes, I'm going to use them even more than before <3 That hero header was really easy to solve with that.
- I didn't spend time on trying to change the ordered list bullets type to "1)".
- Might be a little off-topic, but I searched information about properly using main, section, header tags, even if it's not a big post. Decided to include header inside main and divide main in section tags. I'm really interested in using the html5 semantics in the best way possible, SEO is hungry for it.

PS: Plz, bring apple pies tomorrow. Thx
47rv

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,78 @@
+@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400,600');
+
+body {
+  font-family: 'Open Sans', Arial, Helvetica, sans-serif;
+  font-weight: 300;
+  font-size: 10px;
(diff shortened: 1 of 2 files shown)


ta-web-mad:
Hola Matt.

Flex es una maravilla cuando lo conoces.

Ten cuidado al utilizar la notación #rrggbbaa, ahora mismo es mejor utilizar rgba que es más compatible con navegadores.

Has hecho un gran trabajo Matt, sigue así.

Juan.

PR #734, opened 2018-08-24
MIA ftwd alex rojas

alexrojas:


MarcosPalaciosIII:
good work on the recipe clone. Just remember when linking images from an image folder to use the pathing method so that people that don't have access to your pc can have the image actually show. This is important when you create websites and you have images in the project folder that you will be linking to the site. 

Keep up the good work.

PR #749, opened 2018-09-30
[PAR-PT] JustineZahiri

justinezahiri:
Lab completed.
@ta-web-paris

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,5 +1,7 @@
+@import url("https://fonts.googleapis.com/css?family=Open+Sans:300,600");
+
 body {
-    font-family: Helvetica,Arial,sans-serif;
+    font-family: "Open Sans",Helvetica,Arial,sans-serif;
     font-size:11pt;
 }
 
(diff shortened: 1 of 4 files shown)


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,63 @@
...
+    font-weight: normal;
+    font-family:Helvetica, Arial;
+    color: #ffffff;
+    }
+
+#myTitle {
+    background:url("../images/apple-pie.jpg");
+    background-size: cover;
+    padding-top:350px;
+    padding-bottom:350px;
```
ta-webpt-par: That's not really a good practice to use `padding` to set the height of an element, it would have been better to use a `<div>` containing the `<h1>` and to set the `height: 700px;` for the `<div>` and to set the position of the `<h1>` using `margin`.

But your code works so that's okay !

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,63 @@
...
+    margin-bottom:30pt;
```
ta-webpt-par: It's always better to use the same unit for a web page, you could have stick to `px`.
We will see more about this when we will talk about responsive web pages.

ta-webpt-par:
Good work !

PR #765, opened 2018-10-11
Berlin Axel-Heinz

axelheinz:


axelheinz:
Almost done with the apple recipe part - still have problems formating images: 
a) to position the instruction images and size them equally
b) to scale the header image to the right size

I was also not sure how to make the horizontal line below the sections - I made it now with a margin at the bottom of a box element that I wrapped around the part above the line.

axelheinz:
Done with the apple pie (with some help from Sarah :-).

axelheinz:
I could not find out how to change the bullet points in <ul> items into "-". I tried 
```ul {
  list-style-type: "-";
} ```
but it did not work :-/

PR #778, opened 2018-10-16
Paris - Mathis & Laura

laurastromboni:


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,74 @@
...
-1/2 cup packed brown sugar
-8 Granny Smith apples - peeled, cored and sliced
-
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
+prizes in local competitions. I hope it becomes one of your favorites as well!</p>
+  </div> 
+
+<div class="content image">
+  <img src="./images/recipe-info.png" alt="Image" style="height:70px">
```
ta-web-paris: It's better to keep all the css is the css file. It helps to keep things organised and helps prevents conflicts later which is important when the files get more complex. You can add class names if you need to. 

The alt tag should be something descriptive eg 'recipe info' to provide extra info incase there is a problem with the src link or the user is using a screen reader.

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,74 @@
...
+
+<div class="content image">
+    <img src="./images/cooking-info.png" alt="Image" style="height:100px">
+</div>
+
+<hr class="content"> 
+
+<div class="content ingredients">
+    <h2>Directions</h2>
+    <ul>
```
ta-web-paris: Rather than hardcoding the numbers, you should use an ordered list. This will be helpful later when you're dealing with more items.

```
<ol>
<li>content</li>
<li>content</li>
</ol> 
```

starter-code/french-onion-soup-starter-code/css/main.css
```diff
@@ -7,3 +7,64 @@ html {
   font-family: 'Open Sans', sans-serif;
   font-size: 18px
 }
+
+header{
```
ta-web-paris: If you added a header to your html you would be able to put all your css in one file since both files are almost the same.

```
header .soup {
}
```

starter-code/french-onion-soup-starter-code/index.html
```diff
@@ -1,43 +1,77 @@
...
 
-Heat the oven broiler.
+<body>
 
-Ladle soup into oven safe serving bowls and place one slice of bread on top of each
-(bread may be broken into pieces if you prefer). Layer each slice of bread with a slice of
-provolone, 1/2 slice diced Swiss and 1 tablespoon Parmesan cheese. Place bowls on
-cookie sheet and broil in the preheated oven until cheese bubbles and browns slightly.
+  <header>
+    <h1>French Onion Soup</h1>
```
ta-web-paris: you could add a class here 

```
 <header class="soup">
    <h1>French Onion Soup</h1>
</header>
```

which would allow you to add all css into one file

starter-code/french-onion-soup-starter-code/index.html
```diff
@@ -1,43 +1,77 @@
...
+  <div class="content image">
+    <img src="./images/recipe-info.png" alt="Image" style="height:70px">
+  </div>
+  
+  <hr class="content"> 
+  
+  <div class="content ingredients">
+  <h2>Ingredients</h2>
+  <ul>
+  <li>- 1/2 cup unsalted butter</li>
```
ta-web-paris: if you wanted to get the "-" in front of each li there's a cool way to do it with css (that you'll learn later)

```
.ingredients li::before {
  content: "- ";
} 
```

it adds a "- " before each item meaning you don't need to add them

starter-code/french-onion-soup-starter-code/index.html
```diff
@@ -1,43 +1,77 @@
...
+  
+  <div class="content image">
+      <img src="./images/cooking-info.png" alt="Image" style="height:100px">
+  </div>
+  
+  <hr class="content"> 
+  
+  <div class="content ingredients">
+    <h2>Directions</h2>
+    <ul>
```
ta-web-paris: It should also be an ordered list here

ta-web-paris:
cool! You did this well! 

Remember to use semantic tags when possible because they have features that we can use. It looks great 🙌

PR #779, opened 2018-10-16
Paris priyanka helen

priyaprafful:
@hwoldu

### DIFF
--- a/starter-code/apple-pie-starter-code/index.html
+++ b/starter-code/apple-pie-starter-code/index.html
@@ -2,34 +2,59 @@
 <html lang="en">
 <head>
   <title>Apple Pie Recipe</title>
+  <link rel="stylesheet" type="text/css" href="./css/main.css">
 </head>
 <body>
-Apple Pies
-This was my grandmother's apple pie recipe. I have never seen another one
-quite like it. It will always be my favorite and has won me several first place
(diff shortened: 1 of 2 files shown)


starter-code/apple-pie-starter-code/index.html
```diff
@@ -2,34 +2,59 @@
...
-1/4 cup water
-1/2 cup white sugar
-1/2 cup packed brown sugar
-8 Granny Smith apples - peeled, cored and sliced
-
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
+  <header>
+      <div class="container">
+          <img src="./images/apple-pie.jpg" width="1200px" height="800px">
```
ta-web-paris: For this, you would have been better to set the image as a background in the css file. 

```
header{
  background-image: url(../images/apple-pie.jpg);
} 
```
Then you can add extra styling to position the image. You can adjust the height of the header, the[ background-position ](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position) , and the [background-size](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size) to give you exactly what you're looking for. 

it's also important to note that css should be kept in the css file. This will help prevent conflicts later especially when you're dealing with bigger files

PR #780, opened 2018-10-16
Paris - Harnit & Corrado

Harnitks:
@griev04

We have just done the Apple Pie exercice.


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,71 @@
+body {
+    font-family: 'Roboto';
```
ta-web-paris: to remove the white space from around the header, you can set a margin to 0 here. It's often a good idea to set the margin to 0 because it helps prevents conflicts later on
```
body {
    font-family: 'Roboto';
    margin: 0;
}
```

ta-web-paris:
great work! 

you did well! It looked good and your code was easy to follow 🎉

PR #810, opened 2018-10-22
Miami Full Time Web Development - Oscar Pachano

opachano:


ManuelCabrera01:
Oscar good job  you are only missing small details and the biggest one is making the main image fit the entire div

PR #820, opened 2018-10-23
MIA-FTWD-Yoli-ApplePie

YoliYoli:


ManuelCabrera01:
well done...! the only detail is the path of you main image remember to use a relative path and not a absolute path for the next time

PR #843, opened 2018-10-24
AMS Ana Kelly

anakellyc:


IronhackAMS:
He Ana! Looks good overall:)!

One remark, your cooking info png's are adjusted too much relative to their parent container when the screen size is adjusted. You can fix this by giving it a fixed height:)

PR #844, opened 2018-10-24
AMS Dea Kristina

Deakristina:


starter-code/apple-pie-starter-code/index.html
```diff
@@ -2,34 +2,68 @@
...
-3 tablespoons all-purpose flour
-1/4 cup water
-3 tablespoons all-purpose flour
-1/4 cup water
-1/2 cup white sugar
-1/2 cup packed brown sugar
-8 Granny Smith apples - peeled, cored and sliced
+  <section id="section-heading">
+  <div class="main-header">
+    <h1>Apple Pies</h1>
```
IronhackAMS: the h1 is not centered in the middle:) Try to fix this by deleting the width in your `.main-header h1` and adding `display: flex` and `justify-content: center` to your `.main-header`

IronhackAMS:
Mrs Dea,

Apa kabar?

Pekerjaan yang dilakukan dengan baik.

Beberapa komentar kecil:

- Your cooking info png's are adjusted too much relative to their parent container when the screen size is adjusted. You can fix this by giving it a fixed height.
- Your headings also has some responsiveness issues when you set it to a small screen size. You can fix this for example by giving a rem value in the body or html element (= root element) and then give rem values to different element which then adjust relatively to that root initial value. See: https://www.sitepoint.com/understanding-and-using-rem-units-in-css/ for further info:)

Yours expected,
Kido

PR #845, opened 2018-10-24
AMS Matt Hamers

mattviy:
finalized html&css exercise of day 1 week 1

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,60 @@
+body, html, main {
+    margin: 0;
+    color: grey;
+}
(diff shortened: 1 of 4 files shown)


IronhackAMS:
Mr Matt,

Looks good overall:)
A couple of small remarks:
- Your cooking info png's are adjusted too much relative to their parent container when the screen size is adjusted. You can fix this by giving it a fixed height
- Your h1 also has some responsiveness issues. You can fix this for example by giving a rem value in the body or html element (= root element) and then give rem values to different element which then adjust relatively to that root initial value. See: https://www.sitepoint.com/understanding-and-using-rem-units-in-css/ for further info:)

Yours expected,

Kido

PR #846, opened 2018-10-24
AMS- David Z - Complete CSS Recipes clone

zlobizlobi:


IronhackAMS:
ZlobBRAA,

Looks good!
Some small remarks:
Your cooking info png's are adjusted too much relative to their parent container when the screen size is adjusted. You can fix this by giving it a fixed height.
Your headings also has some responsiveness issues when you set it to a small screen size. You can fix this for example by giving a rem value in the body or html element (= root element) and then give rem values to different element which then adjust relatively to that root initial value. See: https://www.sitepoint.com/understanding-and-using-rem-units-in-css/ for further info:)

PR #848, opened 2018-10-25
Finished

TheHunter896:


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,58 @@
...
+
 
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
-flour to form a paste. Add water, white sugar and brown sugar, and bring to a boil. Reduce
-temperature and let simmer.
+<p id="description">This was my grandmother's apple pie recipe. I have never seen another one
+quite like it. It will always be my favorite and has won me several first place
+prizes in local competitions. I hope it becomes one of your favorites as well!
+</p><br><br><br>
```
IronhackAMS: this is not a clean solution to go onto the next line, try to come up with something else bro

IronhackAMS:
Mr Hunter,

This assignment needs some improvement mate.
-  The website is not yet responsive,  always try to size containers relative to their content and let the children component decide the size of its parent container. You learned about flexboxes, try to implement those with their according property values in your parent containers to make the website responsive:)
- try to avoid using too many id's if it's not necessary. Choose classes over id's when you can.

Yours expected,
Kido

PR #904, opened 2018-11-01
AMS-Luis Wilhelmi Corona

lwilhelmi:


IronhackAMS:
The first site looks good! Try to use class names instead of id's in order make your css multi applicable. In that way you won't have to write any css for the second page (union soup).

PR #917, opened 2018-11-24
[PTMIA1018] - Gustavo Galvez

g-galvez2018:


ta-web-mia:
Looks good! I see you added a max-width to the wrapper which wasn't necessary but it's good to see you experimenting with different techniques :)
- Francisco

PR #944, opened 2019-01-08
Paris - Thomas & Céline

frietpoune:
We learned a bit about Flexbox, next step is to use it properly!


frietpoune:
@tomdreamer

ta-web-paris:
Congrats, it seems almost perfect :)

PR #948, opened 2019-01-08
Paris - Estelle - Alexandre

alexjoly13:
@wnski


ta-web-paris:
Nice :) Some things to improve but that's a very good start :)

PR #981, opened 2019-01-22
Roberto Konanz FTWD - Miami

rxkonanz:


rxkonanz:
First pull request.

ta-web-mia:
Good Job!!! It looks awesome!! There are a couple of things that you could change to meet the best practices:
- Instead of writing 4 lines of code to set the margin
```
  margin-bottom: 40px;
  margin-left: auto;
  margin-top: 40px;
  margin-right: auto;
```

you can do this in one. The fist number would be the top, the second one is the right, the third one is the bottom and the fourth one is the left. 

``` margin: 40px auto 40px auto```
and since the top and bottom and right and left are the same you can also do it with two numbers only
```margin: 40px auto```.
It is ok to use ```margin-top: 40px``` if that is the only margin you want to modify. 

-Since you are repeating 
```margin-left: auto;
  margin-right: auto;
``` 
a lot in several classes you should create one class, let's say 'center' with those properties and you add it to the html tag.

```.center{
margin-left: auto;
margin-right: auto;
} ``` 

``` <img src="images/recipe-info.png" class="cooking-info center"> ```

-Also it is good practice to organize the classes in css from less specific to more specific.
body{}
p{}
h3{}
.my-class{}
.my-class h2{}
#my-id{}

right now you have it like this
html{}
*{}
...
.header{}
...
ul{}
.recipe-info{}
h3{}

PR #988, opened 2019-01-22
nunnapat FTWD-MIAMI

nunna19:
thank you


ta-web-mia:
Good Job! There are a couple of things that you should consider in the future. 
1- For the title section it is better if you wrap the h1 in a div and set the background to the div.
2- Leave the body width as it is and wrap all the content in a big div which could have a max-width of whatever you want it to have and center it with margin: 0 auto.
Other than that, everything looks fine.

nunna19:
Yes 😁 thank you

PR #1013, opened 2019-01-25
PAR- lab-css-recipes-clone // Haydee DONE

haydee75:


Paul-de-Barbeyrac:
Well done !!
- Commenting this `background-attachment: fixed;`will prevent the parallax effect.
- Setting margin left and right to 0 here `margin: 0px 300px;` will give you the same display than in the exercise presentation.
Good job !
`

PR #1065, opened 2019-03-19
Paris arnaud

sunwucode:
Still working on it. Have troubles applying the good properties to tags.


starter-code/apple pie/styles.css
```diff
@@ -0,0 +1,53 @@
...
+    background-image: url(images/apple-pie.jpg)
+}
+
+h1 {
+    color: white;
+    text-align: center;
+    position: relative;
+}
+
+.header_paragraphe {
```
ta-web-paris: The reason these css properties were not applied is because in your HTML you have `id="header_paragraphe"` but here you're calling it as a class. 

so you need to change this to `#header_paragraphe`

starter-code/apple pie/styles.css
```diff
@@ -0,0 +1,53 @@
...
+    top: 0;
+    left: 0;
+    width: cover;
+    
+    margin-block-end: 20px;
+    border-bottom: 2px solid grey;
+}
+
+.header {
+    background-image: url(images/apple-pie.jpg)
```
ta-web-paris: Here, you need to add a height and width attribute to get it to take up more space

ta-web-paris:
Css is hard, and it gets easier with practice! Overall it was good, you just need to be careful with making sure you put a `;` after each css line 🙌

PR #1105, opened 2019-04-01
MIA FTWD APR- Katy Brown/Daniela Olano

katybrown91:


ta-web-mia:
Good job! Be careful about the files that you're editing/adding to the starter code when you fork the repository though. I had to copy and paste the HTML and CSS files outside of the starter code folder to the ones within it. Once I did that though, everything looked pretty good! You've gotten much better at CSS, and the only changes that can really be made are to the font-family and making the page more responsive. Keep up the good work!

ta-web-mia:
Also, the comment about the files being outside of the starter code folder was actually directed at the Spotify Clone lab. Recipes look good!

PR #1107, opened 2019-04-01
BCN 04/19 - Yvonne Moser

YvonneMoser:
Changed Html and CSS


YvonneMoser:
I think i should have added 

margin-right: auto;
margin-left: auto;

to my html selector :)

PR #1108, opened 2019-04-01
Mia FTWD -APR Kenneth & Justin

klsolano7:
done


ta-web-mia:
Good job! Just for reference, I think it would be preferable in the future to create an unordered list using bullets instead of dashes, rather than just utilizing dashes and line breaks. You can also change the style of the bullet point used, but it's nothing too important. For now, focus on the more difficult material ahead. Keep it up!

PR #1111, opened 2019-04-01
MIA FTWD APR - Ryan Kegerreis

ryankegerreis:


ta-web-mia:
Good job! Just a reminder that you can use background-image: cover in order to maintain the resolution of a background image within a section. Keep it up!

PR #1112, opened 2019-04-02
MIA FTWD APR - Chris Cardelle

ccardelle:


ta-web-mia:
Apple pie page looks good! I wouldn't get hung up on not finishing the onion soup page, since you've clearly demonstrated a strong understanding of the HTML and CSS fundamentals on the apple pie page. Stay focused on what's ahead, and keep up the good work!

PR #1113, opened 2019-04-02
MIA - FTWD - Jon 

jonriquer:


ta-web-mia:
Apple pie page looks good! Don't worry about the onion soup page; you definitely have a solid grasp on HTML/CSS fundamentals, so I don't think focusing on it over the more recent material is really worth your time. Stay focused on what's ahead, and keep up the good work!

PR #1114, opened 2019-04-02
MIA FTWB Anghy Pomar

apomarn:


ta-web-mia:
You're on the right track! I think the main thing you need to be careful about is how you're pathing to your images. Right now this is what you have:

`  background-image: url(/Users/AnghyPomar/Desktop/Ironhack/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/apple-pie.jpg);
`

on my end, nothing is showing up since our directories are completely different. Instead, path to the image like this:

`  background-image: url(../images/apple-pie.jpg);
`

It's much shorter, and now it works no matter who's looking at the file. Keep in this mind moving forward, and keep up the good work!

PR #1115, opened 2019-04-02
MIA - FTWD - Nour Cherif Essoussi - APRIL 2019

Nourcs:


ta-web-mia:
Apple pie and onion soup page both look really good. I don't think there's too much to improve on in the basic HTML and CSS department, so stay focused on the more difficult material ahead. Keep up the good work!

PR #1116, opened 2019-04-02
MIA FTWD - Christian Khawam

CodeByCK:


ta-web-mia:
Good job! One small issue I ran into twice though was how you pathed to both the CSS, and the background image within the CSS. When you need to go back a directory in order to access something, it's notated with a dot. This is what you had for the link tag that contained the CSS:

`  <link rel="stylesheet" type="text/css" href="/css/main.css">
`

but since you're going back a directory (since you can't path forward from the HTML file), here's what you should have:

`  <link rel="stylesheet" type="text/css" href=**"./css**/main.css">
`

It might not seem like a huge difference, but this is the difference between your CSS taking effect and it not being connected to the HTML at all. Definitely something to keep in mind moving forward!

PR #1119, opened 2019-04-02
MIA FTWB-APRIL Parker Cooper

parkcoop:


ta-web-mia:
Well done! It doesn't make much of a difference in a page like this, but it's typically better practice to use a background image in the CSS for the top section instead of placing the text on top of an image using absolute positioning. I know I mentioned it earlier as well, but it's an important enough concept that it bears repeating! Everything else looks good though, and I'm glad you've got a solid grasp of fundamental HTML/CSS material. Keep up the good work!

PR #1120, opened 2019-04-02
MIA-FTWD-Allison

allisonkopp:


ta-web-mia:
Looks great, no complaints here! I'm glad you have a good grasp of the HTML/CSS fundamentals already. Stay focused on the more difficult coursework ahead, and keep up the good work!

PR #1121, opened 2019-04-02
MIA FTWD Samuel

YamamotoSamuel:
CSS and HTML updates

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,38 @@
+*{
+    border: 20px;
+    margin-top:0;
+    margin-left:5px;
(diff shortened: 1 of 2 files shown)


ta-web-mia:
You're on the right track! The thing that seems to be tripping you up the most is relative pathing; just remember that going back a directory is notated with a dot, and going forward is notated with a forward slash. So in other words, going back one directory to access your CSS would be notated as the following:

`  <link rel="stylesheet" type="text/css" href="./css/main.css">
`

This will be very important moving forward when you're linking a lot of different files to your HTML, so definitely commit it to memory as soon as you can. If you need a refresher on any of this, feel free to pull either of us aside for a little help. You're on the right track, keep working at it!

PR #1123, opened 2019-04-02
MIA-FTWD- Angel Biondi

AngelBiondi:


ta-web-mia:
Good job, you're definitely getting the hang of HTML/CSS stuff! Stay focused on the more difficult material moving forward, and keep up the good work.

PR #1157, opened 2019-05-28
Paris - Yacine Berrada

yacinebe:


ta-web-paris:
Nice work but you didn't need de back spaces on both size and to add a "–" on a list you can just write it directly inside the html ... but it's a detail :)

PR #1159, opened 2019-05-28
Par Gregoire lescuyer 

gregoireles:


ta-web-paris:
Be careful adding a padding , max height and background cover as they will add up to the background-image.
Try to use flexbox it'll make your life easier !

PR #1160, opened 2019-05-28
PAR ftwd Virginie

Virbee:


ta-web-paris:
perfect ! but you didn't need to add the black background on both side, I think it was just the way it appears on your screen. If it's not clear ask me :) (mimi)

ta-web-paris:
perfect ! but you didn't need to add the black background on both side, I think it was just the way it appears on your screen. If it's not clear ask me :) (mimi)

PR #1164, opened 2019-05-28
Paris FTWD - Camille C. Lab 1 Recipes

camcharp:


ta-web-paris:
Good work, and you did both of the exercises so it's really good ! Plus I notice that you improve in the second exercise. The result is closer from the original design. You can improve on positioning element with new infos that guillaume is giving today

ta-web-paris:
Good work, and you did both of the exercises so it's really good ! Plus I notice that you improve in the second exercise. The result is closer from the original design. You can improve on positioning element with new infos that guillaume is giving today
--

PR #1166, opened 2019-05-28
PAR ftwd Elise

Elisegam:


ta-web-paris:
Good Job ! You are just missing the margin on the <hr /> elements, if you follow the same principle as your sections, it should align as wanted : )

PR #1167, opened 2019-05-28
Paris Camille J

dividemeby0:


ta-web-paris:
Nice work, there are some improvements you could apply by removing the default margin of the body. And in a real world you would use classes or ids to select element in a more specific way

PR #1197, opened 2019-06-08
MIA-PTWD-Alexander Hoinville

ahoin001:


ta-web-mia:
Nice work. Your 'directions' section of the apple pie clone could've been a lot wider, but good job overall.

PR #1199, opened 2019-06-08
MIA PTWD Dusty Garon

dustygaron:
Don't mind the browser reset :)


ta-web-mia:
Structure looks good overall. I'm not seeing a background image in the header at the beginning. I see it in your CSS, but it doesn't appear to be loading correctly. Possibly a problem on my end. Nice work!

PR #1201, opened 2019-06-08
MIA PTWD Antquanique Dancy

adancy95:


ta-web-mia:
Very nice work on the apple pie clone! I see you used font awesome, which is awesome :) I'm marking this completed.

PR #1203, opened 2019-06-08
MIA PTWD Christian Alvarez

calvarez049:
Apple Pie - First assignment - HTML-CSS


ta-web-mia:
Excellent work on the apple pie clone. However, it isn't centered on very large screens. You could center everything by adding the css property "margin: 0 auto;" to your "main-func" class. I'm marking this as completed.

PR #1205, opened 2019-06-08
MIA PTWD Jesus Garcia

JesusGarcia23:


ta-web-mia:
Looks good, but your title image wasn't centered on the page. You can center sections with the css property "margin: 0 auto;" Nice work overall.

PR #1206, opened 2019-06-08
MIA PTWD Oscar Leon

OscarLeon15:
Hello, second time doing the exercise, added one fancy css property.


ta-web-mia:
Very nice, I really liked the fixed background title image effect. Good job.

PR #1207, opened 2019-06-08
MIA-PTWD-Nelly Gonzalez-Vinas: Completed apple pie example.

nellygvinas:
Hi, I did not finish the dashes on the UL, and had trouble with the alignment on the recipe images. Did a workaround with margin and padding. 

MIA-PTWD: Nelly Gonzalez-Vinas.


ta-web-mia:
Looks good overall. If you want to go back now that we learned the **display: flex** properties, you could align things pretty easily, but we aren't requiring it. I'm marking this as completed.

PR #1209, opened 2019-06-08
MIA PT Lab assignment Carlos

carlosdh1985:


ta-web-mia:
This needs a little more work. Now that we've gone over the **display: flex** properties in class, this should be easier. I'm marking it as completed, but needs improvement, so please try to at least get the background image working for the title sometime in the next week or two.

PR #1222, opened 2019-06-10
MIA_FTWD Taryn Chiang

tarynchiang:


MarcosPalaciosIII:
Good job on the apple pie section. For the header div when manipulating a background image you have several tricks you can use with CSS like [background-size, background-position, etc]. CSS can be tricky at times but with more practice, it will become much easier.

Keep up the good work.

PR #1223, opened 2019-06-10
MIA-FTWD-Artem

arstrel:


MarcosPalaciosIII:
Excellent work on the recipe clone lab. You seem to have a good grasp of CSS, though it can seem tedious early on, it will become easier later the more you learn about it.

Keep up the excellent work.

arstrel:
Thank you kindly 😊 keeping the momentum going

PR #1225, opened 2019-06-10
MIA-FTWD-SERGIO VELIKOPOLJSKI

sergiovelo17:


MarcosPalaciosIII:
Good job on the apple pie section. For the image div with the apple pie background image you could have made it max width and then add equal padding on left and right to help get it more centered. For the images, it's usually best to adjust either the width or height but not both. CSS can be tricky at times but with more practice, it will become much easier.

Keep up the good work.

PR #1226, opened 2019-06-10
MIA-FTWD-MARIOSALVATIERRA

mesalvav:


ta-web-mia:
Nice job! I like the workaround you used instead of the images, but definitely practice inserting  tags into the HTML and formatting them properly. Overall, I think it looks good. Keep up the good work 👍 
- Nick

PR #1227, opened 2019-06-10
MIA-FTWD- Andre Cabrera 

andregcab:


MarcosPalaciosIII:
Good work on the French Onion Soup section. You seem to have a good grasp of CSS, though it can seem tedious early on, it will become easier later the more you learn about it. 

Keep up the good work.

PR #1229, opened 2019-06-10
MIA-FTWD-PabloVazquez

PabloVazquezFayat:


ta-web-mia:
Looks pretty good, but there is no "Apple Pie" title over the picture. Also, move the Ingredients section further to the left. Fix these two things, and it'll look great!

PR #1231, opened 2019-06-11
Miami-FTWD-Serena Dominguez 

SDominguez21:


MarcosPalaciosIII:
Good job on the apple pie section. For the header div you can always adjust the width and height of the div in order to have the background image scale more to the size that your looking for. CSS can be tricky at times but with more practice, it will become much easier.

Also keep in mind that you don't have delete a folder if your not going to work on it for submission.

Keep up the good work.

PR #1233, opened 2019-06-11
MIA-FTWD-<Mitch>

mitchmalinin:


ta-web-mia:
Your layout looks good! There were some issues linking other files to the index.html folder. This is a common mistake, and takes some time to get used to, but rather than linking to a file like this:

/starter-code/apple-pie-starter-code/images/apple-pie.jpg
it would be something like this:
../images/apple-pie.jpg

each of the dots represent a previous directory, and when you submit code, chances are it won't be read properly because the directories are not laid out the same (it happened when I tried to view it)

Other than that, it looks pretty good. Just get used to doing this, but it's kind of a muscle memory thing

PR #1234, opened 2019-06-11
MIA-FTWD-Nicolle

nicollegonzalez:


ta-web-mia:
Pretty good! I know you're focused on getting everything centered, but it isn't completely necessary to reduce the width of the header and sections so much. I would suggest adding some kind of margins to the left and right instead, to push everything just a little bit. If you have extra time after finishing all of the current labs, you can go back and improve this. Otherwise, stay focused on the current labs. Keep up the good work 👍 
- Nick

PR #1236, opened 2019-06-11
MIA-FTWD - Michael Roth

mikerawth:
Did submit pull request, but forgot name.  Thanks Marcos


MarcosPalaciosIII:
Great work on the recipe clone lab. You seem to have a good grasp of CSS, though it can seem tedious early on, it will become easier later the more you learn about it.

Also remember not to close your pull request. Once they are checked, then we close them .  :)

Keep up the great work.

PR #1239, opened 2019-06-11
MIA-FTWD Sams Diana 

sdiana11:


ta-web-mia:
This looks good! One thing to look at is how you included your images. Your path was "C:\Users\fahad\Desktop\IronHack\lab-css-recipes-clone\starter-code\apple-pie-starter-code\images\recipe-info.png"

In this instance, I don't have access to your C: drive, so I can't access the pictures. You need to link them like this:

./images/recipe-info.png OR /images/recipe-info.png depending on where your files will be.

Other than that, which is pretty common, good job!

PR #1240, opened 2019-06-11
MIA-FTWD Matiny Louis

Matiny:


ta-web-mia:
Nice work, not much to critique on my end. Really happy to see you're comfortable using sass files, as there's so much more you can do with it compared to a normal CSS. Keep up the good work, and let me know if you have any questions 👍 
- Nick

PR #1241, opened 2019-06-11
MIA-FTWD- Candido Concepcion

candidocon:


ta-web-mia:
Nice work! I would suggest reducing the width of the divs containing the text, and aligning them to the center using the "text-align" property in the CSS. The font color and size could use some minor adjustments, but there's certainly nothing major that needs to be changed. Keep up the good work 👍 
- Nick

PR #1244, opened 2019-06-12
MIA-FTWD-Gulfiya Isyangulova

GulfiyaIsyangulova:


ta-web-mia:
Nice job! Just a reminder, but there was one other image file that had directions on it that was meant to go above the ingredients. Not a big deal though, and I think the way you formatted and positioned elements on your page was very well done. Keep up the good work 👍 
- Nick

PR #1246, opened 2019-06-13
MIA - FTWD Matthew Larrubia

mlarrubia:


ta-web-mia:
Nice job! It looks good on full-size resolution, which is about as much as you can hope for on your first lab. Just a reminder that vh and vw measurements can lead to a more responsive webpage than just using px. Looking back at your first lab, it's pretty amazing how much you've managed to improve in just one week. Keep up the good work 👍 
- Nick

PR #1297, opened 2019-08-07
PAR-FTWD-AIDA

AidaBej:


ta-web-paris:
Nice work, you could improve it by working on how center element in css

PR #1304, opened 2019-08-13
[BER-FTWD-André] - Apple Pie Recipe

itstheandre:


angminsheng:
Checked. All good. em can be replaced with rem for easier responsive design.

itstheandre:
Makes sense. I thought about rem actually. But I was not 100% sure if it would be right solution seeing I didn’t have 100% certainty if it would break the code, because it would inherit sizes from other tags and then grow exponentially.
Thank you for the feedback!

+49 178 493 5615
Linkedin [https://www.linkedin.com/in/itstheandre]
You can already listen to the new episode of the Pioneers Show here [https://open.spotify.com/show/556GnaAw1m54rGyUGJQk8K?si=7YLzCUbBRfCA0HTpwtQyoQ] !
On Tue, Aug 13, 2019 at 19:22, angminsheng <notifications@github.com> wrote:
Checked. All good. em can be replaced with rem for easier responsive design.

PR #1305, opened 2019-08-13
[BER-FTWD-082019] - Jan Redmann

janRedmann:


angminsheng:
Checked. All good. The three sections selectors could be replaced with one class since they are applying the same margin.

PR #1306, opened 2019-08-13
[BER] Moritz Wright 

MoCode:
I have done the LAB


angminsheng:
checked. All good. Clean code and good naming of classes and id.

PR #1308, opened 2019-08-13
[BER-FTWD-NinetteAdhikari]

ninetteadhikari:


angminsheng:
Checked. Looks good. There is one problem with the header section. 
`h1{
    background-image: url("/starter-code/apple-pie-starter-code/images/apple-pie.jpg");
    background-size: cover;
    text-align: center;
    color: white;
    padding: 250px;
    margin:0;
    font-size: 4rem;
}`

This is not a very good practise. The background image should be placed inside a `div` tag and the `h1` should be inside the `div`

ninetteadhikari:
thats good to know 👍🏼

PR #1309, opened 2019-08-13
[BER WD - Iouri Sorokine]

iourisorokine:


angminsheng:
checked. All good. Font family can be applied to the body if it is used everywhere.

PR #1310, opened 2019-08-13
BER-FTWD-Nelly

nellycschneider:


angminsheng:
checked. All good. The word "styling" can be excluded from the naming of the classes and ids since they are most of the time used for styling.

PR #1311, opened 2019-08-13
[LIS-FTWD-Marcos] #156

MakoGS:


GTCarmona:
Hey Marcosss,

Overall You did a great job!! However I just want to point it out, where you have your lists, you have in this format:

`<p><li> banana </li></p>
`
I suggest do it the other way around, ideally the <p> tag is for text only. In this scenario it didn't bring you any problems, but in the long run, it is better to do it the proper way


`<li> <p>banana </p></li>
`

Keep up the good work and any problem let me know

PR #1312, opened 2019-08-13
[BERLIN-FTWD-Marko]

makaveli1313:


angminsheng:
Checked. All good. the naming of the classes for the containers can be done in a more descriptive way instead of container-1, -2 and -3. You can try something like container-top, container-instructions etc..

PR #1313, opened 2019-08-13
BER-FTWD-Keighley

keighleymcf:


angminsheng:
checked. Good job. classes naming like `#cookingInfoPic` can be replaced with `cooking-info-pic` since the naming convention of css classes and ids usually use dashes to seperate the text.

PR #1314, opened 2019-08-13
created html and css

alikucukavci:


angminsheng:
checked. All good. The naming of the classes section-1 , -2 and -3 can be more descriptive, like section-header , section-instructions and etc.

PR #1315, opened 2019-08-13
Berlin-FT-13.08.2019-Nuno.G

nunogervasio:


angminsheng:
Checked. Good naming of classes and well seperated css sections. There is also an option to use rem for the margin and padding which is useful for responsive design. Other than that, good job!

PR #1316, opened 2019-08-13
LIS-FTWD-SamEttinger #156

settinger:


soniameller:
Hey Sam, saw you could do both examples. Great! Just let us know if you need some extra exercises during the first labs if you feel they are too easy for you.

### About the lab:
You are setting a height to the background and padding to the title. This creates a weird behaviour as you can see in the following print screen:

Captura de Pantalla 2019-08-14 a la(s) 09 46 11

By setting the title padding, the parent is talking the necessary height:

Captura de Pantalla 2019-08-14 a la(s) 09 46 50

The rest of it looks very nice, maybe a minor suggestion would be to set a height to the info images instead of a width.

PR #1317, opened 2019-08-13
[BER-FTWD-082019] - Thuy L.

1000miles:
This PR is part of the challenge `Apple Pie Recipe`.


angminsheng:
checked. Clean css structure. hero class can be changed to something like main-container for more descriptive class name.

PR #1318, opened 2019-08-13
[LIS-FTWD-TiagoSilva] #156

TigerSilver:


soniameller:
## General
- Remember to set the margins to 0 so that you don't have the white borders!

## Header
- The font seems a bit small, you can set a different size by adding to the class:
`font-size: "xx"`

## Content
- You found a solution to give some space to the text with the .recipie class, could you think of a way of giving this same solution to the whole content that has the same margin? 
Your solution:
```
.recipe {
  text-align: justify;
  margin: 45px;
}
```

- For the lists, remeber that you have two types. Ordered and unordered lists. The ordered list have numbers instead of bullets. To take it a bit further, check how you can modify the style of the bullets with list-style-type. 
You have some good information in [this link](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type)

-------
Reach me or Gui if you need more help!

TigerSilver:
> ## General
> * Remember to set the margins to 0 so that you don't have the white borders!
> 
> ## Header
> * The font seems a bit small, you can set a different size by adding to the class:
>   `font-size: "xx"`
> 
> ## Content
> * You found a solution to give some space to the text with the .recipie class, could you think of a way of giving this same solution to the whole content that has the same margin?
>   Your solution:
> 
> ```
> .recipe {
>   text-align: justify;
>   margin: 45px;
> }
> ```
> 
> * For the lists, remeber that you have two types. Ordered and unordered lists. The ordered list have numbers instead of bullets. To take it a bit further, check how you can modify the style of the bullets with list-style-type.
>   You have some good information in [this link](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type)
> 
> Reach me or Gui if you need more help!

I ended up only seeing this message after submitting my final but I think every point was corrected. Thanks!

PR #1319, opened 2019-08-13
LIS-FTWD-GuilhermeSãoPedro #156

guilhermespedro:


soniameller:
Hey Gilherme! Here some tips to improve the lab:

## **General**
- Remember to set all the default margins to 0!

## **Header** 
 - In the parent element you could center the text 
```
.header {
  background-image: url("./../images/apple-pie.jpg");
  height: 900px;
  text-align: center;

}
```
- The image looks a bit too big in comparison to the text size. Maybe something near to 400px would have a better proportion
- One posible solution to give some space to the title could be adding some padding to the .title class, you could also use this class to define a bigger font.

## **Main** 
- In the example the text is not aligned to the center so your class for introduction is not necessary!
- If you analyze the example it looks like everything under the header has the same space, a possible solution would be to wrap main and the 2 articles in a div with a class, for example .container and assign it a margin.
- Following the logic you had in the articles, the image could be part of the same container that the text, in this case, you could move the image to be inside the main tag. Check if with the new things you saw in class you can find the way to center it!

## **Articles** 
- You are giving the separation lines to the images classes. That is limiting the line to the image size. Looking at the bigger picture it looks like just the 2 articles have a border in the top. 
- Play around with paddings for the space between elements!
-------

This are all possible solutions but you can try to solve it in many different ways! Reach me or Gui if you have any particular doubt 🙂

PR #1320, opened 2019-08-13
LIS-FTWD-BrandyMcArthur #156

bdmcarthur:


soniameller:
Good work! Just a comment about the solution you used for positioning the title in the header.
When you set a padding to the h1, this will make its parent, with the background image, take that space so it's not necessary to set also a height to the header.
```
header {
    background-image: url('./../images/apple-pie.jpg');
    background-repeat: no-repeat;
    background-position: center;
    background-size: cover;
    height: 800px;  /* --> This could be erased and it would still work! */
}

h1 {
    color: white;
    font-size: 6em;
    text-align: center;
    padding: 325px 0;
}
```

PR #1322, opened 2019-08-13
LIS-FTWD-PaulaBrinkhege #156

PaulaMarcella:
just FYI: i have accidentally started french onion soup instead of apple pie but i just styled it like the apple pie for the exercise


soniameller:
## General
- Remember to set all the default margins to 0 so that you don't have any unwanted white borders.

## Sections
- The text shouldn't be centered in the intro section
- The images could be a bit smaller. Maybe in this case fixing a px unit height could be a good solution
- For image positioning. Remember Ze talked in class about inline and block elements? The images are inline so a possible solution would be the following: try yo go line by line to see if you understand every part and reach us in case you don't!
```
.image-style {
  max-height: 60px;
  display: block;
  margin: auto
}
```
- Check out this [link](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style) for bullet styling!

----
Let us know if something is not clear. Good work!

PR #1324, opened 2019-08-13
LIS-FTWD-BeneditaTeles #156

beneditaalmeida:


GTCarmona:
# heeey bebebebenieee!!!

great job! everything looks very well organized.
One tip I want to give is regarding the class names

```
  <section class="ingredients">
    <h1 class="section-title">Ingredients</h1>
```

maybe it would be better to swap? 

```
  <section class="section-title">
    <h1 class="ingredients">Ingredients</h1>
```

I think it would make more sense. Just want to stress that at some point things will get more complex and is good to already start creating things in a way that is easy to identify and, later on, manipulate

Keep up the good work! :)

PR #1325, opened 2019-08-13
LIS-FTWD-AndreAlmeida#156

andreH2M:


soniameller:
Hey Andre! Some tips for your lab:
## General
- I guess you changed the font size of the whole body to make it look proportional with the image. Usually the default font sizes are fine and in this case maybe the image could be a bit smaller.
- Remember we always want to set the default margins to 0 so that we don't have the white borders
- It would be nice to change the` font-family` of the whole file!
 ## Header
- An easy way to center the text in the image would be:
  - To align horizontaly: `text-align: center` in the parent.  in your case, the `<header>` tag
  - To align verticaly: the `<h1>` could have a padding top and bottom with the half of the hight of the image. If you do this, it is not necessary to give the background a height anymore! It will take the `<h1>` tag space. If this is not clear enough please ask me and I can show you some examples.

## Content
- All your divs that come after your header should have some space in the sides. I can think of two posible solutions. You could wrap all your content in a single div with a `class="content"` for example and:
  - Set a with to that content and center it with `margin:auto`(this would make the content have always the same width)
  - Simply give a margin to the content (this would make the content change size when the screen changes size, respecting the margins that you set )
- Your images could look better if you set a height!
- Separation lines could be created using a border for one part of a section container or by using the` <hr/>` tag
- To go a bit further, check this [link](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style) where you can find documentation on how to change the bullets style

-------

Let us know if you have any extra doubts!

PR #1326, opened 2019-08-13
LIS-FTWD-MiguelCarrilho #156

mangelott:


GTCarmona:
hey Miguel!!

Took a look at your exercise and must say it is very well done and the code is really clean!
Only thing I would add is maybe make the h1 font size a bigger, to make it more similar to the original one, but still, great job! keep it up and any questions let us know

PR #1327, opened 2019-08-13
LIS-FTWD-DanielValero #156

damvalero:
Follow instructions during class

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,60 @@
+*{
+    box-sizing: border-box;
+}
+
+
+body{
+    font-family: sans-serif;
+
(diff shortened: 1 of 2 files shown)


GTCarmona:
HEY DANIEEEL!!

Alright, I took a look at your LAB and overall you did a great job!

I want to point it out two improvements  I would make in your CSS clone make it closer to the original

I would put an automatic margin on the whole page, that way the background image would have the same width as everything else. Right know is a bit weird to see the image with a different dimension 
```
*{
    box-sizing: border-box;
    margin: 0 auto;
}
```
Other thing, I would give some margin or padding between your  h4 tags and the lists, right now they are kinda glue to each other, so usually it makes for a better User Interface to have some space around tags


Keep it up and any problems let me know

PR #1328, opened 2019-08-13
LIS-FWD-GustavoMagnago #156

gusmagnago:
this commit is not honest, it was sent by  mistake. I'm almost done, I know that I have to centralize two images and then I'll be done.


soniameller:
This lab looks very good! Some tips to improve it, you could:
- Use a` <strong>` tag  for the headings so that the look bold
- Set a hight to the images
- Change the style for the bullets. You can check [this documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type)

PR #1329, opened 2019-08-13
LIS-FWD-ElyasPopal#156

elyasp:
Made a clone website for apple pie recipes

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,66 @@
+@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400&display=swap');
+
+* {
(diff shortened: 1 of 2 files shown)


GTCarmona:
Heeeeeeeey Elyasssss!!

I took a look at your LAB, and overall everything is looking really good! One tip I want to give just regarding the name of your classes, because at some point you have this
```

<section class="ingredients">
    <h3>Ingredients</h3>
    <div>
      <ul id="ingredients">


```


I would suggest you name it, for example

```

<section class="ingredients">
    <h3>Ingredients</h3>
    <div>
      <ul id="ingredients-list">


```

With time, our projects will get bigger and complex, so I want to make sure you guys start giving names that are easy to identify and manipulate when you need to.

Besides that, great job!!

PR #1332, opened 2019-08-13
[BER_WDFT]Katy_Stadnyk

KaterynaStadnyk:


brudolce:
Ok, html well structured. During the next labs you can try using flex-box. You did not have this lecture at the time of the lab, so it is all fine.

PR #1334, opened 2019-08-13
LIS-FTWD-FilipeMartins #156

Prelhaz:


GTCarmona:
Hey Filipe,

Something went wrong when uploading to GitHub, there is no CSS file! double check and see what happened, then let me know

thanks!

Prelhaz:
Hello Guilherme,

Ok, I think I know what happened.
The recipe I made was the extra french soup.
Don't know why didn't even noticed the 1st recipe only the extra one...

Just checked in Git and the CSS for the 2nd one is there.

I'll do the first recipe in the weekend, and submit it then, is it
possible? Or too late?

soniameller:
Filipe,

I found your solution. It looks nice! 

In this case it's not important that you did the other one but be sure you read carefully the instructions before you start working so that you don't have problems with more important things in the future.

 You can do the original recipe if you want to get some extra practice but is not mandatory to upload it!

Prelhaz:
Thanks for the heads up, won't happen again for sure.

Have a nice weekend and see you Monday.

PR #1358, opened 2019-08-14
BER-FTWD-KIRAN

kiranshagoti:


brudolce:
Good job Kiran! You are having flex-box lecture now, form now on you can use them to align your html.

PR #1360, opened 2019-08-14
Ber-FT-AUG-Cynthia

salwacynthia:


brudolce:
oke, after today you can try using flexbox ;)

PR #1369, opened 2019-08-14
BER-FTWD-JulianaFerreiraBastos

julferbas:
Styled HTML using CSS, still some issues to fix, like the font, bullets and numbers from ordered lists.


brudolce:
Muito bom Ju! =)

PR #1389, opened 2019-08-19
MIA -FTWD- AUG2019 Kiwi, Andres, Josh

Kiwi-x-Kiwi:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,81 @@
...
+    line-height: 1.5em;
+}
+
+ol {
+    list-style-type: none;
+    font-size: 20px;
+}
+
+ol li:before {
+    content: "-";
```
ta-web-mia: Nice workaround! It's really weird that there's no dashed option for list-styles, so this is the best workaround you can do. Good job!

ta-web-mia:
Nice job! You guys definitely have a solid grasp on fundamental CSS and HTML. Keep it up 👍 
- Nick

PR #1392, opened 2019-08-19
MIA-FTWD-Enrique-Narciso

e-narciso:
French Onion Soup clone


ta-web-mia:
Looks good! Don't forget to put margin and padding to 0 on the entire document to get rid of the small white margin around the webpage.

PR #1399, opened 2019-08-19
MIA-FTWD-ColeMooney&JosueAcevedo

colemooney:


ta-web-mia:
Nice work! Now that you've learned more tools related to responsive design, you should be able to make a page like this look good on multiple resolutions. Keep up the good work 👍 
- Nick

PR #1403, opened 2019-08-19
Lauren L. [MIA-FTWD-082019]

justlask:
first iteration done


ta-web-mia:
Nice job! Just remember to set margin and padding to 0 in the body, to prevent the small white margin on the sides of the screen.

PR #1404, opened 2019-08-19
mia ftwd august Marven Mathelier

marven250:


ta-web-mia:
You're on the right track! I think once you adjust the width of the top section and center the text on the top, you'll be good to go. Keep it up 👍 
- Nick

PR #1405, opened 2019-08-19
webdev Christopher Bolivar

ChristopherBolivar:


ta-web-mia:
Good work! I think the only things that can really be changed are the font-family and some of the measurements. Nothing to concern yourself over though, this is good! Keep it up 👍 
- Nick

PR #1408, opened 2019-08-20
MIA FTWD Jorge 

jorgeluis42:


MarcosPalaciosIII:
Great job on the clone. CSS has a lot o different stylings and can sometimes become overwhelming how many options you have to work with. Just remember to keep things simple when styling and save the more complicated stuff for times that you need it.

Keep up the great work.

PR #1409, opened 2019-08-21
mia ftwd Vitor Tomazzi

VitorTomazzi:


ta-web-mia:
Nice use of background colors to divide the sections. You can accomplish the same thing by putting a border around them as well, but both are good ways of laying out the general structure of a page. As a reminder, make sure you path to files (including images) with relative pathing instead of exact pathing. For example, instead of:

src="/Users/vitortomazzi/Desktop/IronHack/lab-css-recipes-clone/starter-code/apple-pie-starter-code/images/cooking-info.png"

use the following instead:

src="images/cooking-info.png"

Your code was very well structured, so you're definitely on the right track. Keep it up 👍

PR #1410, opened 2019-08-21
mia ftwd Jennifer Muniz & Elba Chimilio

Jenniferrmuniz:


MarcosPalaciosIII:
Goodt job on the clone. CSS has a lot o different stylings and can sometimes become overwhelming how many options you have to work with. Just remember to keep things simple when styling and save the more complicated stuff for times that you need it.

Keep up the good work.

PR #1417, opened 2019-08-22
Dennis Demund Apple Pie

dennisdemund:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,43 @@
+:root {
+    text-align: center;
```
MarcoSantonastasi: Please remove this directive
It is bad practice to align text a global level in a root tag!

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,43 @@
...
+    justify-content: center;
```
MarcoSantonastasi: I am not sure this also belongs in :root

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,43 @@
...
+
+.container {
+    position: relative;
+    text-align: center;
+    color: white;
+}
+
+.centered {
+    position: absolute;
+    top: 50%;
```
MarcoSantonastasi: Very good use of position directive!

PR #1422, opened 2019-08-22
Version Ilies Bourib

IliesBourib-mms:


marianamv112:
Hey @IliesBourib 
Nice lab! Your html is well structured, but be careful with your css, it looks like you submitted it with an error. If you want to improve it, you can try to position the elements more precisely (now we had the flexbox lesson you can try to use it) and use a text font more similar to the example. If you need some help just ask any TA ;) Good work!

PR #1506, opened 2019-10-15
[BER-FTWD-Tatiana-Michael]

crlybrcs:


Svemakawe:
The result looks really good. One tiny thing: if you want the dash instead of the bulletpoints, you can do something like this:
`.ingredients ul {
  list-style-type: none;
}
.ingredients ul li:before {
  content: "-";
  padding-right: 10px;
}`
with removing any default-style from the ul-tag and with the pseudo class "before" you can overwrite the default styling from your list-tags.

PR #1507, opened 2019-10-15
[BER-FT-102019-] - Are-Toska

toskaMaliske:
Excercise apple pie recipe finished by Are and Toska


Svemakawe:
result looks good. I like that you used flexbox for the header. It centers the h1, so you don't necessarily need to have vertical-align on that. For the missing cookingSymbol-pictures, I'll suggest to use again flexbox.

PR #1508, opened 2019-10-15
BER-FTWD-NINO, Shan

NinoGlonti:


Svemakawe:
The recipe-block looks good. The Header could be improved a little bit. I think the h1 tag is not ideal to choose for the background-image. Either wrap your h1 in a div and give that div a background-image, or give the image to your section. H1 should stay a text-element. I like the display flex in your class "apple", you can then align the h1 horizontally with using justify-content: center.
I think it's better to give the body a margin:0 and space the element itself, f.eg. the section.

PR #1510, opened 2019-10-15
BER-WDFT-AlexServie

aslex:


DalinaLeila:
(Comments copied from Jeff's Pull Request): 
Looking good! Maybe play around a bit more with resizing the title (HINT: use font-size in CSS) and resizing instruction images. (HINT: use the width property in CSS).

PR #1511, opened 2019-10-15
[BER-FTWD-Jeff]

Jeff-Moraes:


DalinaLeila:
Looking good! Maybe play around a bit more with resizing the title (HINT: use font-size in CSS) and resizing instruction images. (HINT: use the width property in CSS).

PR #1513, opened 2019-10-15
[BER-FTWD-Fernando&Patrik]

feraraujofilho:


DalinaLeila:
Looks very nice! Next time try to make the naming of classes more specific in order to make it easier for you to look back and understand your code. (Especially when you will work on bigger projects in the future.) 

E.g instead of calling the class for the h1 tag "banner" maybe call it "recipe-title".  :)

PR #1516, opened 2019-10-15
[MAD-FTWD-Noubar]

noutop:


noutop:
width in the .centerImg moves the hole image to the left

PR #1519, opened 2019-10-15
[PAR WDFT] Nicolas

Empty-n:
Petit problème de police d'écriture sur la recette de l'apple pie peut-être ?
Mais je suis sur que ca va passer 🙃

### DIFF
--- a/starter-code/apple-pie-starter-code/style.css
+++ b/starter-code/apple-pie-starter-code/style.css
@@ -1,0 +1,74 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+header {
+  text-align: center;
+  color: white;
+  background-image: url(../images/apple-pie.jpg);
+  background-size: cover;
+  height: 100vh;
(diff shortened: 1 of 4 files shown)


starter-code/apple-pie-starter-code/style.css
```diff
@@ -0,0 +1,74 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+header {
+  text-align: center;
+  color: white;
+  background-image: url(../images/apple-pie.jpg);
```
ta-web-paris: Be careful: here the link is url("./images/apple-pie.jpg") and it would be the same for all the links as your css files is at the same level as your index.html.

ta-web-paris:
For the font, you can use a "sans-serif" one such like Open Sans (you can find it on google font).
Be careful with the src of images: ".." are used to exit a folder but here you create a new styles.css (by the way, one already exists in css folder) at the same level as the index.html.

PR #1522, opened 2019-10-15
BER-WBFT-Amelia&Kyle apple pie clone

scaram02:
Cloned apple pie site


DalinaLeila:
Nice work! 👍 

If you want: try playing around with resizing the main image (e.g. by using the width property in CSS)

PR #1524, opened 2019-10-16
[PAR - WBFT] Nouhad - Apple pie 

nouhadk:
apple pie finished


starter-code/apple-pie-starter-code/index.html
```diff
@@ -2,34 +2,59 @@
...
+    quite like it. It will always be my favorite and has won me several first place
+    prizes in local competitions. I hope it becomes one of your favorites as well!</p>
+   
+    <hr>
+
+</section>
+
+<section class="section">
+    <h3>Ingredients</h3>
+    <br>
```
ta-web-paris: You shouldn't use <br /> to add a space between elements, you should use padding, br has another meaning :)
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

> The HTML <br> element produces a line break in text (carriage-return). It is useful for writing a poem or an address, where the division of lines is significant.

PR #1526, opened 2019-10-16
[PAR-FTWD] Dorota-aplle pie

dodolota:
This is what I produced...


ta-web-paris:
Almost perfect - one thing, on line 22 and 43, just remove one dot so the route of the image works, example: "./images/recipe-info.png" - because you're only descending one level into the image folder

PR #1532, opened 2019-10-16
[PAR WDFT] Hanna

hanna974:
Done


ta-web-paris:
Good job - make sure your images are centered ("margin: auto 0" would suffice here, no need for flexbox)

PR #1535, opened 2019-10-16
[PAR Marion Dev squad 164]

marionboccaccio:
Hi guys, not sure I made the right title! thanks!

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,25 +1,32 @@
 body {
-    margin:0;
-    height:100%;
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Good job! Couple things though, make sure you size the picture so it takes 100% of the screen (height: 100vh helps --> fills the page whatever the screen size is) - also, your img route line 33 : make sure you put the "./", as it sometimes does not work with just "/" ("src="./images/cooking-info.png"). Otherwise all good :)

ta-web-paris:
For the title, it's almost perfect : "[PAR WDFT] Marion" (if you collaborate, put both names)

PR #1541, opened 2019-10-21
done

NoemieNT:


ta-web-paris:
Great job! Just watch your routes, you are not navigating to the right folder (ex: in your CSS file, you are in the CSS folder ---> so ./images is not accessible without going up a level with another dot --> ../images/apple-pie.jpg). Once your three routes are modified, all good :) -- also please name your pull request "PAR WDFT - Noémie"

PR #1584, opened 2019-10-23
MIA FTWD OCT Juan Grajales

JuanGrajales:


ta-web-mia:
The header with the background looks good! If you have any extra time later, try formatting the rest of the HTML. Keep it up :)
- Nick

PR #1587, opened 2019-10-23
MIA FTWD OCT Robbie Sanchez

robbiesanchez:


ta-web-mia:
Nice job! I suggest that from now on, you move elements using margin or relative position rather than padding. This is likely what added that white space at the top of your page. Not a big deal though, and you overall did a great job with this exercise! Keep it up :)
- Nick

PR #1588, opened 2019-10-23
FTWD MIA - October - Pradeepa Shanmugasundaram

pradeepa24:


ta-web-mia:
Nice job! We'll learn more about responsive design in the next few days, so don't worry too much about how it looks on smaller window sizes. Keep up the good work :)
- Nick

PR #1645, opened 2019-11-19
[AMS-WDFT-112019] Dominik Antunovic

Almightyblob:


elyasp:
Good! Tip: To remove whitespace try to use percentages or vw and vh when you want to cover the background with an image, instead of pixels.

PR #1646, opened 2019-11-19
Ironhack - AMS - WebDev - 201911 - Marc de Hoogh

mdehoogh:


elyasp:
Hi Marc, As we see there is whitespace left and right of the page. Where you wanted to make the image cover the full screen, you set the main parent with class .main to an absolute number : 700px;

All of the children will use this as reference, therefore applying 100% to them will mean 100% of that 700px instead of the screen. 

Easy solution is to change 700px by 100%. And 'font-family' is used to change fonts

mdehoogh:
Thx

Op 21 nov. 2019, om 12:28 heeft Elyas Popal <notifications@github.com<mailto:notifications@github.com>> het volgende geschreven:


Hi Marc, As we see there is whitespace left and right of the page. Where you wanted to make the image cover the full screen, you set the main parent with class .main to an absolute number : 700px;

All of the children will use this as reference, therefore applying 100% to them will mean 100% of that 700px instead of the screen.

Easy solution is to change 700px by 100%. And 'font-family' is used to change fonts

PR #1689, opened 2020-01-06
[PAR-WDFT]-PASCAL-BESSON - ongoing

pascalbes:


ta-web-paris:
// Header
Good usage of header! 
To do a hero image, instead of giving the div (class=apple-pie) a fixed height, what you could also do is give the body and the html a height of 100% (and also a weight of 100%), and then you can adjust every other elements with a percentage.
For example, instead of fixing the div height to 200px, you could fix height = 30% or any other percentage. It helps keeping things clean. But your solution is also valide!
To center the text, instead of using a padding-top, you could apply display=flex to your div, and then justify the content and align the items. Tell me if you have a doubt on how to use it!

// Content
Personnally, I would put all the text content (i.e. which is not the hero image) inside a div, and then style it, with padding/margin, depending on what you want to do. Your solution works tho (using :not(h1)) but while reading it, we don't really understand why you are sparing the h1 :D

// General
I think the font-size is quite small. If you want to style it, you can give the html a default font-size (ex: 16x), and then style other elements using em.

// First list
Here, the "li" should be preceded by a "-" instead of the default bullet pooint. You can do that by removing the default list type, and then adding with the pseudo selector ":before" the content "-" :)

// Second list
It's a small nuance but the "li" numbers should be followed by a parenthesis and not a point. You can use this technique if you want: https://www.w3schools.com/cssref/pr_gen_counter-increment.asp
Don't hesitate if you have a question!

// Conclusion
Good use of HTML tags overall :)

pascalbes:
Hello,
I have made some modifications following your comments, but not sure I used git correctly so you can see it.
Can you confirm ?
Thanks !

ta-web-paris:
Hey! You could remove the "width:100%' to your "main" class, and remove the margin of the h1 title "apple pie", that would avoid a strange space on the top. To resize your main image, you could change the pixels so that the image take the whole screen (or use % or vh: that would be recommanded for hero images). Good work on the lists :)

PR #1691, opened 2020-01-06
PAR-WDFT-0120 - Yannick Bourenane

yannick-bourenane:
I used the same CSS file for both projects (not working with lite-server tho)


ta-web-paris:
Great use of css selectors ! Good job :)

ta-web-paris:
Also for the margin of the body, it is always better to wrap all your tags in a container that will have the margin instead of doing it for each tags.

yannick-bourenane:
Ok thanks for the feedback !

PR #1692, opened 2020-01-06
[PAR WDFT]-PAULC - ONGOING PH

pcarillion:


ta-web-paris:
Good job for the overall look of the website, it looks great but there still some tricks to do :

`*,
*>* {
    font-family: "Open Sans", "Helvetica Neue", "sans-serif";
}`
The * selector means select everything, if you do `*>*`, you mean select everything inside everything which doesn't make sense.

For the font-family, write it without the quotes, it will work with quotes but not on every browsers.

You have a margin on your body, this margin is automatically created by your browser, that's why 99% of the time it will always better to start your css with : 

`body {
margin: 0;
}`

Try to do the soup one taking into account all that, you are almost there :)

pcarillion:
new version

ta-web-paris:
You still have a gap on top of your page, it comes from the automatic margin of the h1 tag.
You could have 2 ways to deal with it, either you apply a 0 margin to your h1 as well : `h1{margin: 0}`, or you could do instead of the `body {margin: 0}` use the `*` selector to remove all the automatics margins : `* {margin: 0}`
Anyway good job :)

PR #1693, opened 2020-01-06
[PAR-WDFT]-FLÉCHON Cécile - Done

kamacess:
20200106 - lab css recipes clone completed \o/

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,61 @@
+* {
+    font-family: 'Open Sans', 'Helvetica Neue', sans-serif;
+}
+
+header {
+    background-image: url(../images/apple-pie.jpg);
+    background-size: 100%;
+    min-height: 1000px;
+}
+
+h1 {
+    color: white;
+    text-align: center;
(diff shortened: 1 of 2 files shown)


ta-web-paris:
// GENERAL
- You should put the link to your stylesheet inside the head tag, and not outside :)
- Be careful to the indentation :) Otherwise, your code would not be easy to read. Every element inside another should be indented. Ex: "li" are inside the "ul" so they should be a tab deeper than ul.
- Apart from that, your page looks really nice!

// HEADER
- The strange small margin you see around your image (top/left/right) is actually the body/html default margin. I would advice you to give the body and html a margin of 0 every time you start a project, so you start on a clear basis :) 
- To center the text, instead of using a padding, you could use, on the containing div, display:flex, justify-content:center, align-items: center. 

// ABOUT STYLING
Usually you don't give all the divs a padding. You add one containing div (ex: "content") (and actually you have one div containing all the sections, but you don't make any use of it) and then apply the padding to this div.

// LISTS
Awesome for the styling of the second list :). For the first one, there is a small nuance tho: we have to replace the bullet points by "-". You can do that by removing the default styling and using the pseudo-selector :before, to add a content before the "li"

PR #1695, opened 2020-01-06
[PAR-WDFT]-Guillaume-Marcellin ongoing

guiohm432:
Work advancement at the end of Day 1.


ta-web-paris:
Good job! You could have centered elements below the picture a bit more but that's almost perfect

PR #1697, opened 2020-01-06
[PAR-WDFT] Alix - ongoing

AlixdeR:


ta-web-paris:
Some things to change: the image is supposed to take the whole screen (then you can scroll and see the text), the font-size is too small, and you should center the elements below the image. It won't take long before it's all good though :)

AlixdeR:
modification done

ta-web-paris:
Perfect! :)

PR #1698, opened 2020-01-06
[PARIS-FT-102019] - Mehmet CO. - ongoing

MehmetCan75:


ta-web-paris:
// HEADER
- You could remove the odd margin around your image: it's the default margin of the body. Just remove it by setting it to 0 :)
- Great use of flexbox :)
- For the height: instead of fixing a fixed height, you could give the body/html a height of 100%, and then fix the height of the header to 100%. This way, it will adjust to the size of your screen


// SECTIONS
- You are repeating a lot of margins for sections (ex: section 1 and section 3 have exactly the same code :)). You could either write your css differently (ex: section1, section3 { ... }), or even giving one class to the containing div of all three sections, and then give this div a padding. This way, you don't have to write the class 3 times in your html
- Awesome for the second list and the trick for the numbering. For the first one tho, you should change the bullet points in "-". To do that, you can use the pseudo-selector ":before"


// TO IMPROVE
- You shouldn't use a "hr" to make visually a line. "hr" has a semantic meaning. You should use border-top / border-bottom on your sections
- You should center the two icon images! :)

Don't hesitate to tell me if you have any question :))) Good work!

PR #1699, opened 2020-01-06
[PAR-WDFT]-NINA ONGOING

Gorlanova:
Ongoing progress !


ta-web-paris:
Your code is super clean, that's awesome!

It's a good practice to comment your code as you did it in your CSS :)
As improvement points, I don't see a lot but here are my advices:
- As you are targeting just one existing header here, you don't need to add a class "header", you could target it with the tag selector ( header {...}). While on the contrary, your class "section" is useful since you are applying different styling to different sections. That's great :)
- You don't need the div in your section Intro, since you are not using it for any purpose. The paragraph is enough. Usually "br" is used when there is a significant meaning to the break line (ex: in a poem), and if you can you can try and style this paragraph instead (ex: by playing on its width. But this is really details :D)
  - You could style a "hr" (here you can use it because there is indeed a break in meaning between the different sections) but you could also style the sections, by adding them a border-top/bottom (depending on what you want to do)

  That being said, your work is awesome ! :)

Gorlanova:
Thank you very much ! 
I’ll keep these in mind for next time :)

PR #1700, opened 2020-01-06
[PAR-WDFT]-Kathleen D.

kathd:


ta-web-paris:
Your work looks great :)
It's cool that you also put an effort in finding a font too :)
It's also good practice to remove the default margin from the body, as you did.
To give you an additional solution: you could have used flexbox in order to center the h1 title, by applying display:flex, justify-content: center, align-items:center to the parent div.
Sincerely, I don't have much to say because it's awesome! Congrats :)

PR #1701, opened 2020-01-06
[PAR][WDFT][012020]  - TAYSSIR ELYOUNSI apple pie modification - ongoing

taytay81:
can you please review my pull request i have submitted ? 
thank you

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,38 @@
+.headerapplepie {
+  background-image: url(../images/apple-pie.jpg);
+  background-size: cover;
+  height: 50vh;
+  position: relative;
+}
+h1 {
+  color: white;
+  text-align: center;
+  padding-top: 20vh;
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Could you please edit the title of your pull request to "[PAR-WDFT] - your name" (if your part of the paris web dev full time team).
Thanks.

ta-web-paris:
So far so good, few things I have noticed : 
- When you are creating classes and Ids, try to make them as specific as possible, for example "text" won't work on big projects.
- Also try to use either caps or "-" to separate when you have multiple words on one class, example "sectionstyle" would be "sectionStyle" or "section-style", (I personnally use caps but you do as you want, just be consistent).
- Don't forget when you do some css, 99% of the time you won't need the automatic borders of the browser, so to avoid that you can do : `body {margin: 0}` (Your h1 tag has an automatic margin so you can as well apply it to your h1).

You are almost there, keep up the good work :)

taytay81:
thank you for your comments , updated accordingly .

PR #1702, opened 2020-01-06
[PAR-FTWD-0120] - Remy Pham - ongoing

kurosaek:


ta-web-paris:
Looks great ! Bravo !
- Just don't forget to do `body {margin: 0}` on your css to remove the automatic borders of the browser.
- Try to create classes to avoid applying css directly to html tags (especially with the div tag that is used a lot).

Good job.

PR #1703, opened 2020-01-07
[PAR - FT- WD] Jonathan Oreja ongoing

Jarchiii:


ta-web-paris:
I think you placed the index.html and main.css files you worked on in the wrong folder (not inside the starter code > apple pie). Delete the ones that are just plain text, and place them in the right folder. If you can't see the picture appearing on the page, it means the path isn't correct. Otherwise the styling is pretty good, you might need to center the items more though + include the main title picture :)

PR #1704, opened 2020-01-07
[PAR-WDFT] - Pierre Etryhard ongoing

PierreEtryhard:


ta-web-paris:
Your work looks great! 
The only improvements I see are only details but here are my advice just in case:
- For the first list, instead of adding manually the "-" before the li, you could remove the default bullet points, and then use the pseudo selector ":before" to add the property content: "-"
- For the second list, in order to have the right formatting "1)" instead of "1." you could use a counter: https://www.w3schools.com/cssref/pr_gen_counter-increment.asp
But these are only details :) 
Don't hesitate if you have questions. Otherwise, you work is great!

PR #1705, opened 2020-01-07
[PAR-WDFT] ALEX MEYER ongoing

ameyerfr:


ta-web-paris:
Almost good! You'd need to change the path for the images: "../images/cooking-info.png" (you're going one level up in the folders, only need one dot before the / --> "./images/cooking-info.png". Also the main picture should take up the whole screen, but I'll close the pull request since these are details. Come ask us if you don't understand!

ameyerfr:
> Almost good! You'd need to change the path for the images: "../images/cooking-info.png" (you're going one level up in the folders, only need one dot before the / --> "./images/cooking-info.png". Also the main picture should take up the whole screen, but I'll close the pull request since these are details. Come ask us if you don't understand!

For the images path, well spotted, my bad :) About the main picture i dont know what you mean, because it does take up the whole screen width.

PR #1706, opened 2020-01-07
[PAR-WDFT] DICKERSON - ongoing

pauldkn:


ta-web-paris:
Good job !
Don't forget to add `body {margin: 0}` to your css to avoid having the automatic margins of the browser.
And I would have done the images a bit bigger but overall it looks great ! :)

PR #1707, opened 2020-01-07
[PAR-WDFF]-baptiste WARGNY- ongoing

baptistewar:


ta-web-paris:
Hey! Good works :) Here are some of my comments:

// General
- In your HTML, you need to write the tags in small letters :)
- It was not very clear on the lab but the hero image should take the whole screen (on width). Therefore, you don't need the property background-size: contain
- You should remove the default padding and margin of the body (by setting it in your CSS: body {padding:0, margin:0}), to get rid of strange space :)

// Your HTML
Semantically, you are using the right tags!
Here, you can use the "hr" to make a line because there is indeed a break in meaning between the different secitons. What you could also do is adding a border-bottom or border-top to your sections (it's another way to do it)

// Your CSS
- In your header, you are using some properties that you don't need: flex-wrap: nowrap; align-content: center. Come and see me if you want us to review some of them together :). Also, I can show you a way to target 100% of the screen instead of fixing a height manually if you want!
- You are not using section2 or section3, so you can remove them 
- To center the recipe-info and cooking-info, what you could do is wrap the images in divs, and then apply margin:auto to those divs. That would enable you to remove the recipe-info class, which is quite complicated, and to remove the property "text-align:center" on section 1 (which is why the first image is centered, and not because you are using margin:auto in your class "info")
- Does the "-" on your ul works? Because it does not work on my computer. Otherwise, you also have this solution :to change the type of "li" in ul, and replacing the bullets points by "-" you should remove the default style (on the ul with list-style:none), and then adding the custom "-" with li:before {content: "-"}
- There is a tip to style the second list! You could use this method if you'll like: https://www.w3schools.com/cssref/pr_gen_counter-increment.asp

Recap: good work but try to simplify some of your classes and methods, especially the ones regarding the images cooking-info and recipe-info. Come and see me if you have any question

baptistewar:
modified according to the comments

ta-web-paris:
Top pour les modifs :)

PR #1708, opened 2020-01-07
[PAR-WDFT]-MORSLI - ongoing

PaulineLalou:
quelques difficultés avec le tableau complexe


ta-web-paris:
Good job !
Few things for you to improve : 
- You have on your page 3 different font families, one on h1, one on the text-intro and arial the automatic one on the rest of the page. It is always better to keep one font-family for your all website for consistency purposes, you can still choose multiple ones but it needs to be for specific style reasons.
- When you select multiple font families on one element : `font-family: Trebuchet MS, Lucida Sans Unicode, Lucida Grande, Lucida Sans, Arial, sans-serif;` it will always be the first one that is taken into account so the other ones are obsolete.
- Don't forget to add as well `body {margin: 0}` or `* {margin: 0}` so it will remove the automatic margins on your document (for example the white borders all around your page).
- For your classes and the alt of your images, try to be consistent as well with the language you are using, the website is in english but some classes are in english and some other in french, the alt is in french so it will display a french sentence in a english web page.

Try to apply that either on the apple pie either on the onion soup. After that comment the pull request so we can have a look :)

Keep up the good work !

PR #1709, opened 2020-01-07
PAR WEB-DEV FT-Youval ongoing

YuvalBu:


ta-web-paris:
Great job! The only thing would be to make the main picture (with title) take up the whole screen, but it's a detail so I'll close the pull request :)

PR #1711, opened 2020-01-07
[Ber-FT] Tim-Mia

timbotimber:
Make us some pie

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,50 @@
+body {
+  font-family: "Montserrat", sans-serif;
+}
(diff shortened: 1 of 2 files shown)


itstheandre:
Hey!
The only thing I have to say about this is that the .header div had a huge margin above.
That's because all elements usually have a default value of padding and margin. If you want to remove them you can either hard code it in the elemnt something like this:

```
.header {
...
margin: 0;
padding: 0;
}

```

or introduce a defaulter in the begining of the css file.

something like this:


```
*{
margin: 0;
padding: 0;
}

```

however you will learn more about this during the week :)

timbotimber:
Thanks Andre!

PR #1712, opened 2020-01-07
FT WD Berlin Jan/2020 - Alexis, Mohamed

Perdikal:


itstheandre:
Nice Job!
Nothing much to comment. 
Maybe you could have tried and making the images bigger (the ones with the infos)

itstheandre:
Ah, I had forgot. Please pay attention to how you source the files. 
The images had a path that was not relative to the css and html files. In your computer it might work, but when this gets shipped it should always be shipped with a path relative to the original file

PR #1713, opened 2020-01-07
[BER-FT-012020] - Johnny Burke, Jacqueline Chen

JohnnyAL:
Done


itstheandre:
Great lab!
Interesting seeing that you've used the ::before pseudo-class. Its not very much used (im not very comfortable with it for example)

Besides that, the header still had a little bit of padding and margin.
You should try and pay attention to that because most elements have a default internal value for those and you might need to take them out

chenjacqueline:
Thanks for your feedback, André!

What is your preferred way to manipulate the bullet points / numbering list styles?

PR #1714, opened 2020-01-07
[BER-FT] Jason - Elsa

ElsaFil:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,53 @@
...
+  position: absolute;
+  top: 37%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  color: white;
+  font-size: 4em;
+  font-family: Arial;
+  font-weight: 300;
+}
+
```
fonsogms: I would suggest  avoid using h1 as much as possible as it is better practice to use a class instead.

ElsaFil: didn't know that, thanks!

PR #1715, opened 2020-01-07
[BER-FT] Jonathan - Marcel

Marsu-R:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,73 @@
...
+body {
+  margin: 0;
+  font-family: ‘Source Sans Pro’, sans-serif;
+  font-weight: 100;
+  font-size: 14px;
+  color: #333;
+  width: 100%;
+  height: 100%;
+}
+
```
fonsogms: Nice design, looks good, I would have added padding to the body so it looks a bit more like in the image. Try adding something like this in the body ```padding: 0 120px```

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,73 @@
...
+h1 {
+    font-size: 48px;
+    font-weight: bolder;
+}
+
+h2 {
+    font-size: 24px;
+    font-weight: lighter;
+}
+
```
fonsogms: I would try to avoid using h1 or h2 as much as possible, as it will better practice to use classes instead.

PR #1716, opened 2020-01-07
BER_FT_Konstantin_Akua

akups:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,58 @@
+@import url("https://fonts.googleapis.com/css?family=Verdana");
+
+body {
+  font-family: "Verdana";
+  margin: 5px;
+  padding: 0 5px;
+  width: 80%;
+  height: 80%;
+}
+
```
fonsogms: Design looks nice!
All good!
Small tip here try to give some more padding to the body so it is more centric in the webpage

PR #1717, opened 2020-01-07
[BER-FT]Sebastian-Sanka

MngSebastian:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+h1 {
+  position: absolute;
+  top: 30%;
+  left: 35%;
+  transform: (-50%, -50%);
+  font-size: 75px;
+  font-weight: 300;
+}
```
fonsogms: All good in the hood!
Design looks nice!
Just small tip, avoid using ```h1{}``` only if neccesary, is always better to use a class first or a class.
Also for the ```# identifiers``` make sure you only use them for very specific sections of a website

PR #1718, opened 2020-01-07
done Kiyohide & Saeka

kiyokiyo2929:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,39 @@
+.headerimage {
```
DalinaLeila: Looks good 💯 Not many comments to give! 
At the moment you have padding from the top of your website, so I'd suggest to take that away to make it more similar to the example. Maybe also use `padding` or `width` to change the div including the paragraph under the header img.

PR #1719, opened 2020-01-07
FT-BERLIN0120 Vaso-Shaul

vasilikibalidou:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,52 @@
+html {
```
DalinaLeila: Not many comments to give, it looks great! 💯 

I would suggest giving classes to tags to target specific items with styling. Also, use `id` mainly when you want to give a name to a section. Otherwise, just use `class`.

PR #1725, opened 2020-01-07
[PAR - WDFT] - Shannon K. ongoing

tranquildoe:


ta-web-paris:
Hey Shannon :)
Your site looks nice :) ! 
Some advice to make it even better:
- you should get rid of the default margin and padding of the body in the css
- to make the image full screen you could get rid of the margin that you put on your class picture, as well as the default margin of your h1 "Apple Pie"
- to center your RecipeInfo image and CookingInfo, you could simplify your code. For example, if I take the id="recipeInfo": you could say that your recipeInfo is display:flex and use the justify-content:center property in order to center whatever is inside, namely the image. Then, you don't need any margin on the img anymore.
- Instead of applying margin-left to several of your elements, what you could do instead is wrapping all the text (i.e. from your class="about" to the end) in a div, and then apply a padding to this div. This would avoid you to have to specify a margin for every text element.
Regarding the HTML itself: 
To have the indentation of your code automatically fixed, there is a option to add in "Prettier". We can do it together if you want:) 
Great progress! Just keep going :) Try to fix the margin thing and it would be great! Don't hesitate to come and see me if you have any question :)

PR #1726, opened 2020-01-07
[PAR-WDFT] - Coline Velard ongoing

colinevel:


ta-web-paris:
Almost perfect, you would just need to make the apple pie picture take up the whole space :)

PR #1727, opened 2020-01-07
[PAR-FTWD-012020] - Maria.MUNIZAGA ongoing

mmunizaga:


ta-web-paris:
Everything looks great, except the last section which could use some margin on the sides, but I think you understood the concepts so I'll close the pull request :)

PR #1751, opened 2020-01-09
[PAR-WDFT] Andrew NB ongoing

andrewnunezbibby:


clarabesset:
Great job !! - one small detail though, you'd need have the apple pie background image of the header to take up the whole screen (before scrolling and seeing the recipe). But you understood the main concepts for this lab so I'll close your pull request :)

andrewnunezbibby:
Thank you! It was hard to tell how much of the screen was supposed to be taken (based on the pdf) so I just sized to the cup lol. I must have missed the instructions somewhere. 

I also just realized that I put li's in the "Ingredients" section, but in the PDF it looks like they might just be regular <p>'s, with numbers at the beginning. let me know if I need to change.

clarabesset:
Yes you're supposed to zoom in so the pdf takes up the whole page but I don't think that was clear either!  But you're all good on the ingredients part, these are details.

PR #1754, opened 2020-01-11
done

fhoman:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,95 @@
...
+
+body {
+
+font-size: 21px;
+
+}
+
+
+
+div.container {
```
lotofcaffeine: Hey, One alternative to make the image bigger would be instead of using a margin, try to give it a width like this:
`header {
  background-image: url("../images/apple-pie.jpg");
  height: 100vh;
  width: 100%;
  background-repeat: no-repeat;
  background-size: cover;
  text-align: center;
  color: white;
}`

lotofcaffeine: Just to let you know, to practice how to use html5 tags, you can use the tag "header" instead of a div with class.

PR #1759, opened 2020-01-13
PARWDFT0120 ANTHONY L . [ongoing]

anth0sama:


starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,57 @@
...
+
+.header {
+    background-image: url("../images/apple-pie.jpg");
+    background-repeat: no-repeat;
+    background-size: cover;
+    position: relative;
+    height: 100vh;
+    padding-top: 380px;
+    padding-right: 120px;
+}
```
ta-web-paris: here you could have played with the background properties:
background-position: center;
background-size: cover;

to make a better background image.

You can take rid of the padding-top and padding-right values aswell as they will add up to the size of the background, therefore not covering 100vh, unless you add box-sizing:border-box     property.

https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

SInce you added the relative position, you could position the text as absolute to center it over the background.
Or you could have used flexbox and it's properties to get the same result.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,57 @@
...
+    font-size: 16px;
+    line-height: 4em;
+    
+}
+
+h3 {
+    color: black;
+    font-size: 14px;
+    text-align: center;
+    padding-top: 50px;
```
ta-web-paris: You coud set the font-weight to 300 so the text appears less thick.
font-weight:300;

https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,91 @@
...
+
+        <li>Place the bottom crust in your pan. Fill with apples, mounded slightly. Cover with a
+          lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
+          that it does not run off.</li>
+
+        <li>Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
+          degrees C). Continue baking for 35 to 45 minutes, until apples are soft.</li>
+      </ol>
+    </article>
+</div>
```
ta-web-paris: You could have wrapped this whole part into a div or section to which you would have applied the padding-left and right properties to create equal space around all your elements.

starter-code/apple-pie-starter-code/css/main.css
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.image img {
+    width: 300px;
+    margin-left: 29%;
+    margin-right: 29%;
+
+}
```
ta-web-paris: here instead of having the margins set to fixed percentages you could display the image as block and have the margin: auto property center the image

.```
image img {
    width: 300px;
  display:block;
margin: 0 auto; // apply equal margin on the left and right
}
```

ta-web-paris:
Be careful selecting directly the tags when styling elements.
Tag selectors will style those elements globally, so if someone tries to update the html he might have some of his tags coming in with styling which is what he may not want.

PR #1784, opened 2020-01-15
[BCN.WEB.HENRIETTE]

Jethet:
It will become better, I am sure ;-). There is something with the background picture: it seems like more of the picture is visible in my clone than in the original in the repo.


tawebbcn:
Very good, it looks very similar!
Great job

PR #1788, opened 2020-01-16
[BCN - WEB - Sara del Valle]

LolaEnBeta:


tawebbcn:
Good in general, try to be more regular for the font-size in your page.
The last image should be more in the center.

LolaEnBeta:
Thank you for the comment! I didn't see it! I create I new pull request before see it with the finished work, and I think that the image you say, now it's in the correct position (I hope). :)
Thank you so much!

PR #1789, opened 2020-01-16
lab is almost done

Tking1606:


tawebbcn:
The exercise looks good! the first list was ordered and the second unordered but besides that it looks ok.

Just remember to do the pull request witn the name [BCN-WDFT-YOURNAME] so we know its you among the other ironhackers around the world ;)

PR #1790, opened 2020-01-16
I was tired, I'm sorry

valentin-muller:


tawebbcn:
- The structure of the html looks okay. 
- In the css if you set the font-family (or any other property) to the body or to everything ( *{} ) you wont need to set it for every section as you did. 
-Also, very important. When you do a pull request, create the title starting with [BCN-FTWD-YOURNAME]

PR #1791, opened 2020-01-16
sorta done

valmgisbert:
I think the basics are there, but I still want to tweak stuff! feedback appreciated!


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,54 @@
...
-Apple Pies
-This was my grandmother's apple pie recipe. I have never seen another one
-quite like it. It will always be my favorite and has won me several first place
-prizes in local competitions. I hope it becomes one of your favorites as well!
+  <head>
+    <title>Apple Pie Recipe</title>
+    <link rel="stylesheet" href="./css/main.css">
+  </head>
+  <body>
+    <h1>Apple Pies</h1>
```
tawebbcn: here you could wrap your h1 element with a <header> tag, just for semantic purposes ;)

PR #1793, opened 2020-01-16
[BCN-WEB-ADI]

JuiceDrinker:


starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,67 @@
...
-8 Granny Smith apples - peeled, cored and sliced
 
-Directions
-Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in
-flour to form a paste. Add water, white sugar and brown sugar, and bring to a boil. Reduce
-temperature and let simmer.
+<body>
+  <div id="jumbatron">
+    <h1>Apple Pie</h1>
+  </div>
```
tawebbcn: Here you could use the < header > tag, better than div in terms of semantics :)

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,67 @@
...
 
-Place the bottom crust in your pan. Fill with apples, mounded slightly. Cover with a
-lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
-that it does not run off.
+  <main>
+    <section class="note">
+      </p>This was my grandmother's apple pie recipe. I have never seen another one
+      quite like it. It will always be my favorite and has won me several first place
+      prizes in local competitions. I hope it becomes one of your favorites as well!
+      <p>
```
tawebbcn: Opening < p > tag and closing tag are inverted by error, so the browser considers the img below is inside a p element.

starter-code/apple-pie-starter-code/index.html
```diff
@@ -1,35 +1,67 @@
...
+        <li> 3 tablespoons all-purpose flour</li>
+        <li> 1/4 cup water</li>
+        <li> 3 tablespoons all-purpose flour</li>
+        <li> 1/4 cup water</li>
+        <li> 1/2 cup white sugar</li>
+        <li> 1/2 cup packed brown sugar</li>
+        <li> 8 Granny Smith apples - peeled, cored and sliced</li>
+      </ul>
+      <img src="./images/recipe-info.png" alt="" class = "small-img" srcset="">
+      <hr>
```
tawebbcn: If you place you < hr > elements outside of sections, they will be both the same height (on the contrary, the sections styles apply to them, and they might be different)

PR #1812, opened 2020-01-22
[LIS JAN20 - Filipe Morais]

FilipeMorais43:
background image not working

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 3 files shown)


soniameller:
Let us know if you can solve it or if you need help! 
 @ironhackLis-webTeam

FilipeMorais43:
I believe that it is working, but I did it with to teacher help otherwise my background-image was all wrong. I am still struggling with images and padding/margins and etc on CSS. 

by the way, I am trying to pull request the last lab (cluedo) and I can´t do it with the terminal. I am missing the last interation and interation 2b, but this last one is (I believe) for small details. Functions confuse me a lot

PR #1841, opened 2020-02-12
LIS-FEB-PTWD Stefano Santonocito

oddouu:
@Ironhack-LIS-PTWD
Should be done! Just a few little notes:
- The ul doesn't have dashes but dots. I tried to research the css property "list-style-type" but probably I wasn't on the right way;
- I was unsure if fixing the height of the pictures cooking-info and recipe-info, but then I left them as they are;
- I first declared a lot of classes but ended up not using them;
- The second list is an ol, but the numbers are separated. I guess that's right!

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -8,13 +8,13 @@ body {
 }
 
 .title {
-  background-image: url("../images/apple-pie.jpg");
-  background-size: cover;
-  color: white;
-  height: 60vh;
-  display: flex;
-  justify-content: center;
-  align-items: center;
+    background-image: url("../images/apple-pie.jpg");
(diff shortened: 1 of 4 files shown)


ironhack-LIS-PTWD:
Hey Stefano,
It's looking good! 

If you want the dashes and not the dots, you can use `list-style-type: none` and add in `::before`
Take a look at https://stackoverflow.com/questions/47327961/html-for-list-with-dashes-and-custom-font-colour and see if you can make the dashes work :) 

If you wanted to change the size of the information images, you can add a class and adjust in the css (but this isn't a huge concern) 

Do not worry if you are not using all the classes for styling, it is better to give classes because you may want to change the styling of certain elements later on and this will be useful for you. 

The ordered list numbers are good :) 

Good job! 👍 
Lissy

PR #1844, opened 2020-02-12
LIS - WDPT - FEB - Wilson Cruz

galatoide:
After class, going halfway.


galatoide:
@ironhack-LIS-PTWD

Done :)

PR #1846, opened 2020-02-12
LIS - PTWD - FEB - Julia Carvalho

juliamendesc:
@ironhack-LIS-PTWD
Could not do the bar below the pictures. Was that supposed to be part of the image or should we have inserted ourselves?
Other than that, I think I am good with this exercise. No further queries. :)


ironhack-LIS-PTWD:
@juliamendesc  Hey well done for completing it 👍  
It's looking good and you've structured your code well and named the classes nicely! 
If you want the line under the image, you can use a <hr> tag which gives a horizontal line break in a page :) 
- Lissy

PR #1870, opened 2020-02-17
BCN-WDFT-VÍCTOR RONDA

victorronda:


victorronda:
I was referring to "margin" when I said "padding" in the commit d956439.

PR #1891, opened 2020-02-18
LIS - PTWD - FEB - Rui Pereira

RuiMPer:
tag @ironhack-LIS-PTWD
working on it, I'm little late on the labs, but until friday I hope I will catch.
 
having difficulties to put "This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always be my favorite and has won me several first place prizes in local competitions. I hope it becomes one of your favorites as well!"  on the center of the page, already did flex direction column and justify on center.

### DIFF
--- a/starter-code/apple-pie-starter-code/css/main.css
+++ b/starter-code/apple-pie-starter-code/css/main.css
@@ -1,0 +1,20 @@
+body {
+    margin: 0;
+    font-family: sans-serif;
+}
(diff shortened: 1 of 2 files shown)


RuiMPer:
@ironhack-LIS-PTWD
almost done.
I don't know what I'm doing wrong on the last image, I can't center it like the second image.

help? ;)

PR #1897, opened 2020-02-20
AMS WEB FT FEB - Justin Peeters

JustDevInn:


mswidzinska:
* Technical Requirements (10/10)
* Creative Code (10/10)
* Code Quality (10/10) 
* Front-end (10/10)
Well Done. Very readable code and clear naming.
Small tip: It is better to use  %  than px for margins

PR #1906, opened 2020-02-21
AMS WEB FT FEB Shabeer Hussain

shabeerhussain:
Moring,


MFn-Scarboy:
Technical Requirements (9/10)
Creative Code (9/10)
Code Quality (9/10)
Front-end (8/10)

Remember to try and use '%' instead of 'px'
HTML indentation could use some work

All in all pretty good
-- Joel

PR #1931, opened 2020-03-23
PAR WDFT Divya - Done

dgopal23:
Hello,
I completed the "Recipes" Lab. Could you please tell me if it is okay?
Thank you,
Divya

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -3,19 +3,45 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Very nice ! 

the only few things I have to tell you :
- the “article” around the “p” in your first section, either you put the image inside it as well, or you can just put “p” (article with only one thing inside is not needed)
- and the big image you can put a bit more height to it (i don’t know if you know about vh unit, it’s very nice because it adapts to the screen of everybody ! you can check it online :wink: )
- and last : the “debug” feature that you will have to put a lot if you want perfect design without the browser messing things up :

html, body {
  margin: 0;
  padding: 0;
}

so that everything is against the border of the window
other wise it’s perfect ! well done !!

PR #1932, opened 2020-03-23
[PAR WDFT] Guillaume de Glisezinski - done

TheShua:
Done (with the wrong font but .. eh)


ta-web-paris:
[Michel] OK : rien à dire.

PR #1933, opened 2020-03-23
PAR WDFT 032020 Sam L - done

samlfair:


ta-web-paris:
[Michel] OK. My tip: nest the <h1> inside a container <div> or <header> and set the image as a background of it, rather than putting all the css settings into <h1>. HTML elements are meaningful, even though your solution's perfectly fine :)

PR #1938, opened 2020-03-23
[PAR WDFT 032020] William Hermozo - in progress

williamhzo:
Apple-pie recipe done


ta-web-paris:
[Michel] OK. As discussed with you, make sure you push a new commit with the slight modifications below:

- Change URLs to make sure images are well displayed on non-local machines
- Change the `background: url(...)` as `background-image: url("...")`

PR #1941, opened 2020-03-23
[PAR-WDTF] - Mathias Gautier - waiting for improvement

saitam1990:


ta-web-paris:
[Michel] OK. Could you please apply the modifications below before I close the PR?

- Add `* {
	margin: 0; /* to make sure the navigator does not apply its default value */
	padding: 0; /* to make sure the navigator does not apply its default value */
	box-sizing: border-box; /* to  avoid calculating the size of your content via calc(...) */
}` at the very beginning of your CSS file.
- Fix the indentation (which the alignment of each line of your code base). One way would be to go to VS code settings (Code > Preferences > Settings > in the "Search settings..." bar, look for 'format on save' and check the associated box)
- Instead of `padding: 300px;` for the h1, prefer `line-height: ` equal to **the same value as the parent container** (i.e. 600px in your codebase)

Keep those fixes in mind ;)

ta-web-paris:
OK. Some other fixes:
- `recipe-infos`: you need to add '.' in the beginning of the src path -> `recipe-infos`. Same goes for the other image.
- Bullet styling: instead of putting '-' in the HTML, check [this](https://developer.mozilla.org/fr/docs/Web/CSS/::before) out and figure out another CSS solution.
- List number styling: check [this](https://css-tricks.com/custom-list-number-styling/) out and figure out another CSS solution. 

If you can't, we can talk tomorrow about it :)

ta-web-paris:
OK. New commit fixed those issues.

PR #1950, opened 2020-03-24
[PAR-WDFT-032020] - Yuval Bugajski Ongoing

YuvalBu:


ta-web-paris:
Nicely done Yuval !

You used rem, % and vh / vmin, that's very good for future responsive designs.

One small remark : the h1 with a background : I would recommend a div wrapping your h1, and the div would have the background (that way if you're asked afterwards to add a paragraph or logo, etc... under or next to the title but still on the image your box stays the same) but in this case it works just fine ! Nice job

PR #1953, opened 2020-03-24
BER-FTWD-Maria-Sami

Alba-Tros:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,99 @@
...
+
+#section1 {
+    display: flex;
+    flex-direction:column;
+    align-items: center;
+}
+
+#section1 p{
+    text-align:center;
+    width:572px;
```
Jeff-Moraes: This width property will be better at the parent element style (#section1)

apple-pie/styles/style.css
```diff
@@ -0,0 +1,99 @@
...
+ul.dashed {
+  list-style-type: none;
+}
+ul.dashed > li {
+  text-indent: -5px;
+}
+ul.dashed > li:before {
+  content: "-";
+  text-indent: -5px;
+}
```
Jeff-Moraes: nice approach!

PR #1956, opened 2020-03-24
BER Recipes Paula & Dioni

dionifss:
Just finnish ;)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,36 @@
...
+  list-style-type: "-";
+}
+
+.little {
+  display: block;
+  margin: 0 auto;
+}
+
+.main {
+  margin: 0 200px 0 200px;
```
fenrew: In my opinion: Instead of applying margin to force something to stay in the middle it can be more beneficial to put the width to something like:
width: 80vw;
This will make it much more responsive

PR #1957, opened 2020-03-24
BER-FTWD-Sebastian&Olha

olhadanylchenko:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,67 @@
...
+h2 {
+  font-weight: 400;
+}
+
+#imgHeading {
+  z-index: 1;
+  color: white;
+}
+
+#headImg {
```
fenrew: Do you actually use this id?

The way you ended up solving the header image is what I at least consider the best way:)

olhadanylchenko:
thank you for your feedback, @fenrew! 
cc @SebastianNagl

PR #1959, opened 2020-03-24
BER - FTWD - Benjamin and Yung 

Benchberlin:


apple-pie/index.html
```diff
@@ -1,21 +1,60 @@
...
-  <body>
-    Apple Pies This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Apple Pie Recipe</title>
+  <link rel="stylesheet" href="./styles/style.css">
+</head>
+<header>
```
fonsogms: The header tag should be inside of the body tag :).
The only things inside of the head tag should be meta data and links to other resources.
Anything regarding content should be inside of the body tag.

PR #1961, opened 2020-03-24
BER-FTWD-Chantel-Timur

HeiseMo:
Done

### DIFF
--- a/apple-pie/styles/style.css
+++ b/apple-pie/styles/style.css
@@ -1,0 +1,114 @@
+* {
+    margin: 0;
+    padding: 0;
+  }
+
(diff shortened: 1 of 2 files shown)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,114 @@
...
+    width: 800px;
+    margin-left: 30px;
+}
+
+.ingredients {
+
+
+}
+
+.ingredients ul {
```
fonsogms: This was a correct way of doing, but if you still have any questions you can still ask Markus and Jeff since they have better experience in design :)

PR #1962, opened 2020-03-24
BER FWD-Daniela

daniela-kirilova:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,69 @@
...
+    padding: 0;
+}
+body {
+    font-family: "Arial";
+    padding-bottom: 50px;
+  }
+header {
+  background: url("../images/apple-pie.jpg") no-repeat;
+  background-size: cover;
+  height: 500px;
```
fenrew: Keep in mind that using vw, vh, % and em will make your code way more responsive. In this case, 500px is fine though as I can see you've been nice and creative for the rest of the code:)

PR #1965, opened 2020-03-24
PAR WDFT 202003 Anne Patana - ongoing

apatana:


ta-web-paris:
Nice job Anne ! I like the layout in the middle of the page

Nice use of margins and line height. Now that we've seen flexboxes, you can use them for your future projects ! With practice you'll se they're very well... flexible haha

Very cool styling of the list to have "-" and "1)" !

A small point to keep in mind for the future :
- Your layout is very nice, but for responsive design (depending on screen size) try to start using %, vh/vw to size elements, and em / rem for fonts, and some margins / padding (if you have any question about them please ask ! you can find tons of articles online (css-tricks best website ever) )

Bravo for your first lab :)

Nina

PR #1966, opened 2020-03-24
PAR WDFT 0320 Teddy ROGERS ongoing

TeddyRod:


ta-web-paris:
Nice job Teddy !
A few points to keep in mind :
- You can put the images in sections (instead of creating another section just for them) that way your code will be shorter and cleaner, and don't put a footer if you don't use it :)

- Nice trick for the lists, you hid the list style and added the - and 1) in the text, but there is a way to change the list type without adding text in the <li> ! You should learn how to do it because it can be exhausting to chance every <li> if you have 2000 of them (and with a few lines of code it does the same job) for example :
>> for the "-" in the first list just type
ul {
    list-style-type: '- ';
}

>> for the 1) in the second list you can look here
https://stackoverflow.com/questions/22530982/html-numbering-with-brackets/22531181

- Nice use of vh and % for element size ! That's what you will be using for future projects ! You can try to stop using px (except for small details) and use em / rem for margin, font size and padding (you can look online and ask me if you have questions) and only % or vh / vw for the rest

Tell me if something is not clear, otherwise good job :)
Try to practice the points I gave you to improve !

Nina

PR #1977, opened 2020-03-29
done

mferret:


ta-web-paris:
Hey Max ! 

Nice looking lab :)
A few remarks for your future code (i do not expect you to correct the lab and send it again since the result is clean, it's just for you to read my feedback carefully and try to keep it in mind for next time)

HTML : 
- please put every content in the "body" section, your header should be in the body ! it does not change the result on most browsers but on internet explorer / old browsers you will have an error.
>> = HTML =
= = HEAD = = 
(here all the meta content, everything that is not visible on the page but is important information, page title, css links, language, unicode versions, scripts... etc)
= = / HEAD = = 
= = = = = = = = 
= = BODY = =
(here all the content of the web page, header, main, sections, div, p, a, ul, etc etc etc... footer)
= = / BODY = = 
= / HTML =

- you can put your "img" directly in the code, no need to put it in a "div" if it's the only element inside it :)

CSS : 
- you don't need to specify "flex-direction: row" and "flex-wrap: nowrap" : they're the default values :)
- the "text-shadow" has only a color value, you have to put width and height if you want to see it : for example "text-shadow: 4px 4px black"
- you should add a small padding for the top of your sections (it will be more "aéré", for instance "padding: 5% 10% 0" will be top = 5, left / right = 10, bottom = 0)

Otherwise nice work !
Hit me up for questions if you need !

Nina

PR #1978, opened 2020-03-29
[PAR-FT-032020] Maxime Paolone

mxpln:
source

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -1,24 +1,74 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey Maxime ! 

Good job ! Very clean :)

- you don't need header + div around your image, you can use the header as the div and it's cleaner
- design wise, i would avise you to put the content (except the hero image at the top) a bit more in the middle (more padding in your wrapper on the sides) to have a more "modern" look 

The rest is very clean, perfect !
Hit me up if you have any question,

Nina

PR #1979, opened 2020-03-29
[PAR-WDTF-032920]Olivier Brulé

OBxel:
Apple pie exercice not totally finished due to not enough time

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -3,19 +3,69 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="./styles/style.css">
     <title>Apple Pie Recipe</title>
-    <!-- don't forget to link your styles -->
   </head>
-  <body>
-    Apple Pies This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always
+
+  <body class="bg-setting">
+    
+    <header class="main">
+      <img 
+      class="img-setting"
+      src="/images/apple-pie.jpg" 
+      alt="apple-pie">
+      <h1 
+      class="mid-text">Apple Pie
+      </h1>
+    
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey Olivier ! 
Your display looks good overall, here are some points for you to read and keep in mind for next time (i don't need you to correct and resend your lab, it's just a feedback for your future projects :) )


HTML : 
- careful about the indentation, computers don't mind but humans who read your code will bother if your indentation is not clean, it's harder to read !
- you should not put a section in a section, it's working but semantically it would be better to put another tag (so many tags for boxes : header / footer / main (to be used only once in the code), section, article, div... from the biggest piece to the smallest ! if you don't know which one, use div it's safe)
-  you can organize your code like this for example (there are many ways but here is one) : 
>> BODY
-- HEADER
---- image
---- title
-- /HEADER
-- SECTION
---- p 
---- ARTICLE
------ image
------ hr (it's a horizontal break, use it for the big line)
------ title
------ ul (with li inside)
---- /ARTICLE
---- ARTICLE
------ image
------ hr (it's a horizontal break, use it for the big line)
------ title
------ ol (with li inside)
---- /ARTICLE
-- /SECTION
/BODY

CSS : 
- i recommend you use a "body {}" at the beginning of your styling to put everything general in it (font family, background color, font color etc etc) and margin / padding to 0 so the browser doesn't put weird white space somewhere
- If you want a more modern look, ditch the beige for white (or something closer to white)
- you can put the image full width of the screen OR less height to keep the aspect ratio of the mock up
- use your "font-family" in the "body" styling, that way it goes everywhere
- you can use this trick for parentheses in the ol : [here](https://stackoverflow.com/questions/22530982/html-numbering-with-brackets/22531181)

Otherwise cool styling !
Nice use of flex, vh and em ! Ready for responsiveness 

Hit me up for any question !

Nina

PR #1980, opened 2020-03-30
[PAR-FT-03/2020] - Benjamin Dean

bdean05:


ta-web-paris:
Hi Benjamin :)

The result is clean but the code can be improved in a few ways, if you can just have a deep look at it for you future projects i don't need you to actually correct it and send it again. But remember these because it's important for your skills : 

HTML : 
- please keep the content in the "body" section, it's the standard way (your div / h1 in "head" is not correct)
- the h2 in the top part of your body is empty, remove it if you don't use it !
- do not use "br" if you just need space, use margin and padding instead, it's more correct semantically and "br" should only be used for line break inside of a text.
- do not put "img" in the "p", and put them afterwards instead, it's different elements. if you want to group the two of them, use a div to wrap them !
- You replaced the beginning of the list lines by " -" but imagine having to do this for a 2000 items list, try to find out a css way to style it !  Same for the "1)" list at the bottom, there is a trick. For this you should use "ul" and "ol" tags, with "li" elements inside them, not a "p" with "br" (it's semantically more correct)
- try not to do inline styling (like width = 350, and without unit ?) if you have a css file, it should go there :)
- be careful with indentation in your code, it does not make a difference for the computer but for humans reading it it's cleaner if it's done according to standard !

CSS : 
- you don't need to repeat "font-family" for headers since you declared it for the whole body section
- you are styling a "table", "th" and "td" style but you don't have those elements in your html, remove it if it's not used !
- in your ".content" section you set a "margin: auto" which puts blank spaces around the image, design wise you can set margin to 0 it will be cleaner and more modern :)

Hit me up if you have questions !

Nina

PR #1981, opened 2020-03-30
Louis Alibert - PAR "articles"

lalib19:
Sorry I didn't finish it I didn't put enough work into it I am very slow at html/css, same thing for my spotify

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -4,18 +4,74 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Apple Pie Recipe</title>
-    <!-- don't forget to link your styles -->
+<link rel="stylesheet" href="styles/style.css">
   </head>
+
+
+
   <body>
-    Apple Pies This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always
-    be my favorite and has won me several first place prizes in local competitions. I hope it becomes one of your
-    favorites as well! Ingredients 1 recipe pastry for a 9 inch double crust pie 1/2 cup unsalted butter 3 tablespoons
-    all-purpose flour 1/4 cup water 3 tablespoons all-purpose flour 1/4 cup water 1/2 cup white sugar 1/2 cup packed
-    brown sugar 8 Granny Smith apples - peeled, cored and sliced Directions Preheat oven to 425 degrees F (220 degrees
-    C). Melt the butter in a saucepan. Stir in flour to form a paste. Add water, white sugar and brown sugar, and bring
-    to a boil. Reduce temperature and let simmer. Place the bottom crust in your pan. Fill with apples, mounded
-    slightly. Cover with a lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey Louis ! 
No panic here, it's okay not to finish as long as you tried your best and we can review it in time to give you feedback :)

I don't need you to rework on it and send the changes, but read these points carefully so it can help you next time when you're writing code : 

HTML
- be careful about the indentation, it does not make a difference for the browser, but a huge one for the human reading it (it's faster to read when well aligned everywhere)
- you don't need to do a DIV > ((HEADER > P) + IMG) you can directly create HEADER > (P + IMG) it will be shorter ! remember DIV is kind of a box, and a very generic tag, and these are boxes too, but more specific : header (div at the top with key information, like a title, or a nav, only used once) / main (big chunk of code representing the main part of the page, only used once) / footer (bottom of the page with additionnal information (only used once) / nav (with links to other parts of the website) / section (big chunk of code, can be used several times, could be pieces of the page) / article (has a format that can be repeated, like blog posts or tweets or instagram posts or online-store item...) / etc etc etc ... So if you can use these, it's best practice because your code will have more meaning (if we read "main" we know it's the main piece, etc) but if you're not sure, use DIV ! (and DIV is for everything else, like wrapping a title and an image to style them together, etc etc) 

- with that in mind, your sections titles (h2) can be inside the article ! 
- for the line, you can use the tag "hr" that means "horizontal rule", it's a ""séparateur"" so semantically it's the best way to create it
- the second article with the instructions should be a list as well (but not a "ul", it should be a "ol" since it's ordered by numbers) and there's a trick to make the parentheses [here](https://stackoverflow.com/questions/22530982/html-numbering-with-brackets/22531181)

CSS
- for your background image, you can use "background size : cover" "background repeat : no repeat" and "background position : center" to make sure your image will behave the same way on all screens (look them up if you don't know them, they're useful !)
- you can use "display : flex" with "justify content: center" and "align items : center" to place your title in the background, instead of line height (it was not required for this lab but now you know flexboxes !) because line height will make you go crazy if you have to add content after, you will have to change the values
- you can specify the font family in the "body {}" at the top of the css file so your font is everywhere, (and in the "body" i recommend to put "margin: 0" and "padding: 0" to be sure the browser doesn't add white space"
- you can use "em" and "rem" units to size your font, look it up if you need, it's best practice because depending on the size of the screen / zoom for disabled people, it will adapt ! (=> responsive design) you can use them for the margins and padding as well (then you think "how many letters long do i want the padding / margin to be ?)

- ".cooking" and ".recipe" have the same properties, you should create a unique class "recipe-img" or "pictograms" or whatever, then you use it for both images, and then you write only one css block :)
- try to use % or vh / vw to size your boxes or sections, same than the font size, it will be responsive ! (it's always % of the parent element, so the size has to be defined on the parent as well, for example a div width 100% with two children width 33%, the two children will take 2/3 of the space of the div) (and vw = vue width / vh = vue height, so that's a % of the browser window size, pretty useful sometimes !)

Otherwise you have a clean code, it's working ! Just these tips to make it even better, and it's a lot but i try to go into details to give you as much info as possible for future.

Hit me up if you have any question !
Your lab is validated :)

Nina

PR #1991, opened 2020-04-01
lab-css-recipes-clone 

RicardVillalba:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,58 @@
...
+  top: 400px;
+  color: white;
+  font-size: 50px;
+  letter-spacing: 0.9px;
+}
+header {
+  width: 960px;
+  height: 900px;
+  background: center / contain url("../images/apple-pie.jpg") no-repeat;
+  margin: auto;
```
tawebbcn, line 19: It seems that "contain" (background-size) is the reason why you have a top margin. To remove it, you could add a negative top margin:
`margin-top: -40px`

apple-pie/styles/style.css
```diff
@@ -0,0 +1,58 @@
+@import url("https://fonts.googleapis.com/css?family=Roboto+Condensed:700|Roboto:100,300,700");
+
```
tawebbcn, line 2: A best practice is to add a reset CSS at the beginning of your stylesheet in order to remove all default browser styles: https://meyerweb.com/eric/tools/css/reset/

apple-pie/styles/style.css
```diff
@@ -0,0 +1,58 @@
...
+  margin: 10px auto;
+}
+
+h3 {
+  position: relative;
+  left: 20px;
+}
+
+ul {
+  list-style: "- ";
```
tawebbcn, line 49: Well done

apple-pie/styles/style.css
```diff
@@ -0,0 +1,58 @@
...
+  background: center / contain url("../images/apple-pie.jpg") no-repeat;
+  margin: auto;
+}
+
+section {
+  width: 600px;
+  border-bottom: 1px solid black;
+  /*margin-left: 200px;*/
+  margin: 10px auto;
+}
```
tawebbcn, line 27: You could also add some line-height here to add more space between text lines
`line-height: 25px;`

PR #1992, opened 2020-04-01
lab-css-recipes-clone Alberto Pagoria WDFT March 20

alpagor:
Did my best even without using the <ul> or the <ol> ones! Next time I'll give them a try!!!

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -1,21 +1,57 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Apple Pie Recipe</title>
-    <!-- don't forget to link your styles -->
-  </head>
-  <body>
-    Apple Pies This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always
-    be my favorite and has won me several first place prizes in local competitions. I hope it becomes one of your
-    favorites as well! Ingredients 1 recipe pastry for a 9 inch double crust pie 1/2 cup unsalted butter 3 tablespoons
-    all-purpose flour 1/4 cup water 3 tablespoons all-purpose flour 1/4 cup water 1/2 cup white sugar 1/2 cup packed
-    brown sugar 8 Granny Smith apples - peeled, cored and sliced Directions Preheat oven to 425 degrees F (220 degrees
-    C). Melt the butter in a saucepan. Stir in flour to form a paste. Add water, white sugar and brown sugar, and bring
-    to a boil. Reduce temperature and let simmer. Place the bottom crust in your pan. Fill with apples, mounded
-    slightly. Cover with a lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
-    that it does not run off. Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
-    degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
-  </body>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Apple Pie Recipe</title>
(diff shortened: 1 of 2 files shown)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,129 @@
...
+}
+blockquote:before, blockquote:after,
+q:before, q:after {
+	content: '';
+	content: none;
+}
+table {
+	border-collapse: collapse;
+	border-spacing: 0;
+}
```
tawebbcn, line 44: Well done for using the CSS reset ;)

apple-pie/styles/style.css
```diff
@@ -0,0 +1,129 @@
...
+
+ .bkgimg {
+    background-image: url("./../images/apple-pie.jpg");
+    background-size: 90%;
+    background-repeat: no-repeat;
+    background-position: 70px;
+	padding-top: 200px;
+	padding-bottom: 200PX;
+	padding-right: 200px;
+	padding-left: 200px;
```
tawebbcn, line 62: Here, instead of writing 4 padding rules, don't hesitate to use the shorthand padding property (sets the padding on all four sides of your element):
margin: 200px;

apple-pie/styles/style.css
```diff
@@ -0,0 +1,129 @@
...
+}
+
+
+
+#presentation {
+	text-align: left;
+	padding-top: 50px;
+	padding-left: 200px;
+	padding-right: 200px;
+}
```
tawebbcn, line 87: same here, you can use the shorthand padding property, writing: `padding: 50px 200px 0` (top/right/bottom/left, no need to specify the left if it's the same than right, or the bottom if it's the same than top)

apple-pie/index.html
```diff
@@ -1,21 +1,57 @@
...
+    <br>
+    <p>
+      - 1 recipe pastry for a 9 inch double crust pie<br>
+      - 1/2 cup unsalted butter<br>
+      - 3 tablespoons all-purpose flour<br> 
+      - 1/4 cup water<br>
+      - 1/2 cup white sugar<br>
+      - 1/2 cup packed brown sugar<br>
+      - 8 Granny Smith apples - peeled, cored and sliced<br>
+    </p>
```
tawebbcn, line 39: I know you know, but here a < ul > would be more adapted 😉 
To have the "-" bullet style, you can do 
`ul li:before {
  content: "- ";
}
`

apple-pie/index.html
```diff
@@ -1,21 +1,57 @@
...
+  <section id="directions">
+    <h2 class="titulo">Directions</h2>
+    <br>
+    <p>
+      1)Preheat oven to 425 degrees F (220 degreesC). Melt the butter in a saucepan. Stir in flour to form a paste. Add water, white sugar and brown sugar, and
+      bring to a boil. Reduce temperature and let simmer.<br> 
+      2)Place the bottom crust in your pan. Fill with apples, mounded slightly. Cover with a lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
+      that it does not run off.<br> 
+      3)Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175 degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
+    </p>
```
tawebbcn, line 53: Same here, you could use a < ol > list. The CSS would look like this:
```
ol {
  counter-reset: list;
}

ol > li:before {
  content: counter(list) ") ";
  counter-increment: list;
}
```

apple-pie/styles/style.css
```diff
@@ -0,0 +1,129 @@
...
+
+#recipe {
+	padding-top: 20px;
+	width: 370px;
+	height: 60px;
+	padding-left: 300px;
+	padding-right: 200px;
+}
```
tawebbcn, line 95: A bit more dynamic way of centering the image (the risk of using padding with px value is that it doesn't look centered on all screen sizes) is to place the img inside a < div > element and apply flexbox on it (div being the flexbox parent of the img). The CSS would look like this:

`
div {
    display: flex;
    flex-direction: column;
    align-items: center;
`

PR #1994, opened 2020-04-01
[BCN-FT-ALEX A.] - done

skanndar:


tawebbcn:
Great job! Nice use of flexbox, as well as ids and classes.
Some comments below :)

PR #1995, opened 2020-04-01
Apple pie exercice

arimagic-8bit:
Good evening! 

I finished the Apple pie lab. I hope it's ok! :D 

Thanks!! 

Ariadna

### DIFF
--- a/apple-pie/styles/style.css
+++ b/apple-pie/styles/style.css
@@ -1,0 +1,72 @@
+h1 {
+  color: white;
+  background-image: url(/apple-pie/images/apple-pie.jpg);
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: 100%;
+  height: 250px;
+  text-align: center;
+  padding-top: 200px;
+  font-size: 70px;
+}
+
+p {
+  font-size: 22px;
+  text-align: center;
+  width: 50%;
+  margin-left: 500px;
+  margin-right: -300px;
+  letter-spacing: 0.5px;
+}
+
+header img {
+  margin-left: 41%;
+  width: 400px;
+}
+
+hr {
+  width: 80%;
+}
+
+.infoText img {
+  width: 350px;
+  margin-left: 40%;
+}
+
+.infoText {
+  margin-left: 100px;
+}
+
+.infoText h2 {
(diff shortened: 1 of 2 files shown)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,72 @@
...
+
+ul li::before {
+  list-style-type: none;
+  content: "-";
+}
+
+ol li::before {
+  counter-increment: list;
+  content: counter(list) ")";
+}
```
tawebbcn, line 72: well done for finding this one! 👍

apple-pie/styles/style.css
```diff
@@ -0,0 +1,72 @@
...
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: 100%;
+  height: 250px;
+  text-align: center;
+  padding-top: 200px;
+  font-size: 70px;
+}
+
```
tawebbcn, line 13: A good practice would be to add at the beginning of the stylesheet a set of reset rules, in order to remove all default browser styles (this is why you have a top margin on your header for example).
You can use the one from Eric Meyer: https://meyerweb.com/eric/tools/css/reset/

apple-pie/styles/style.css
```diff
@@ -0,0 +1,72 @@
...
+p {
+  font-size: 22px;
+  text-align: center;
+  width: 50%;
+  margin-left: 500px;
+  margin-right: -300px;
```
tawebbcn, line 19: Here you are relying on the margin left and right to center your header paragraph (and same for the image "recipe-info"), so it looks good on a big screen but not so much on smaller screens. 
You could place those 2 elements in a < div > for example and apply flexbox on it in order to center them in a more reliable way :)

apple-pie/styles/style.css
```diff
@@ -0,0 +1,72 @@
...
+  margin-left: 170px;
+  width: 30%;
+}
+
+.infoText ul {
+  margin-left: 140px;
+  list-style-type: none;
+  font-size: 20px;
+  letter-spacing: 0.5px;
+}
```
tawebbcn, line 52: Same comment here, using flexbox instead of playing with margins will help you to get all your elements centered on all screens :)

PR #2054, opened 2020-06-02
WDFT June, Berlin, Ivan Mickovski and Leo Costa

imickovski:
Gave a basic try, need to try to find solution how to do the dashes before the lis in the ul, and to make the lis in the ols with 1), 2), 3).


imickovski:
Done!
Still struggling cannot solve the last ol, to be like: 1), 2), 3)

PR #2059, opened 2020-06-02
WDFT-BER-062020

lupocordero:
Lupo and Nina WD

### DIFF
--- a/apple-pie/styles/style.css
+++ b/apple-pie/styles/style.css
@@ -1,0 +1,85 @@
+@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@300&display=swap');
+
+/* main{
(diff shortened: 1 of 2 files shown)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,85 @@
...
+
+.aligned-elements {
+	display: flex;
+	flex-direction: column;
+	flex-wrap: nowrap;
+	justify-content: space-around;
+	align-items: center;
+	align-content: space-between;
+}
+
```
Jeff-Moraes, line 67: you can add it to limit the size of the images inside the .aligned-elements div
`.aligned-elements img {
  max-width: 100vw;
}`

PR #2090, opened 2020-06-11
done

Menhaj01:


TatijanaRajcic:
Hey Menhaj! Thanks for your lab! Some remarks:
- Be careful, you are not closing the ol list right: you wrote "0l" so it's not closing properly
- Please install Prettier on autoSave so that everytime you save, it formats your code properly: it's really life changing!
- You don't need the max-width on the paragraph class, it creates an artificial space at the right on bigger screens (such as mine), whereas you already center it properly with the property text-align:center
- You don't use your .recipe-info class, so just remove it from your CSS
- You use twice the selector img.cooking-info whereas every selector should only appear once in your css: keep in mind that with this..
```
img.cooking-info {
  margin: 50px auto 0;
  max-width: 280px;
  display: block;
  text-align: center;
}

img.cooking-info {
  max-width: 400px;
  padding: 50;
  margin: 50px auto 0;
  max-width: 380px;
  display: block;
}
```

... the second occurence of img.cooking-info will override the first so it does not make sense

- The property "align-items" goes with flex: if you don't use flex, then it's not useful to use this property either (ex: on your ".recipe-info" class)
- Have a look at [this property ](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type)to customize your bullets points

- Have a look at [this property ](https://www.w3schools.com/cssref/pr_gen_counter-increment.asp)to implement a counter before your ordered list

Keep up the good work and don't hesitate if you have any question :muscle:

PR #2100, opened 2020-06-29
[BCN-FT-062020] Luis Manuel De Gouveia Sousa

luisdegouveia:
I think this is a little bit similar to what we were asked to do in the exercise. I did the HTML and CSS. I need to learn more about HTML and CSS.


tawebbcn:
Buen trabajo :) 
El ancho y alto de las imágenes de cooking info se podrían agrandar un pelín porque se ven un poco "estrujadas", pero el resto se ve bien ;)

PR #2105, opened 2020-06-29
[BCN-FT-062020] Ferran Puig

puigmar:
Create layout from base


puigmar:
apple-pie & french-soup layout

tawebbcn:
Excelente trabajo, te felicito!
Capu:)

PR #2131, opened 2020-07-05
[PTAMS062020] Mahmut Sarcan

mhmtdl:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,101 @@
+.container {
+    width:90%;
+    margin:auto;
+}
+
+.image {
+background-image: url("../images/apple-pie.jpg");
+width:65%;
```
guido-an, line 8: According to the design provided the image should have more width

PR #2133, opened 2020-07-05
[AMS - PT - June20] - Yuka

yukajyeh:


apple-pie/index.html
```diff
@@ -1,21 +1,60 @@
...
-    all-purpose flour 1/4 cup water 3 tablespoons all-purpose flour 1/4 cup water 1/2 cup white sugar 1/2 cup packed
-    brown sugar 8 Granny Smith apples - peeled, cored and sliced Directions Preheat oven to 425 degrees F (220 degrees
-    C). Melt the butter in a saucepan. Stir in flour to form a paste. Add water, white sugar and brown sugar, and bring
-    to a boil. Reduce temperature and let simmer. Place the bottom crust in your pan. Fill with apples, mounded
-    slightly. Cover with a lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
-    that it does not run off. Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
-    degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
+    <header>
+      <h1>Apple Pie</h1>
+      <img id='applepie' src ='./images/apple-pie.jpg' alt='apple pie'>
```
guido-an, line 15: A better way of approach this would be to apply with CSS the image as background of the <header>. Then you can simply use _text-align_ and _margin-top_ to give the correct position to the <h1>

apple-pie/styles/style.css
```diff
@@ -0,0 +1,68 @@
...
+#applepie {
+    height:400px;
+    width: 750px;
+    margin-bottom: 35px;;
+}
+
+
+.first-section{
+ text-align: center;
+  margin-left: 50px;
```
guido-an, line 34: You could also achieve the same goal in one line: `margin: 0 50px`. The 0 is referring to _top_ and _bottom_. The 50px to _left_ and _right_

apple-pie/styles/style.css
```diff
@@ -0,0 +1,68 @@
...
+    display:block;
+    margin-left:auto;
+    margin-right:auto;
+}
+
+#cooking-img{
+    width:40%;
+    height:auto;;
+    display:block;
+    margin-left:auto;
```
guido-an, line 50: Same as above

PR #2136, opened 2020-07-08
[WDFT-REM-2007] YOUR NAME

brittahelm:
definitely agree that css can be more annoying that javascript, feel like i keep not even seeing things so i hope i didn't forget anything :D


IH-WebDev-TA-Remote:
Looks perfect! very good job :)
Only little bit missing is the last list to do the ")" instead of the "." but still looks amazing!
For the next pull request, please add your name to the title where it says "your name".

PR #2137, opened 2020-07-08
[WEB-REM-2007] RAFAEL DE LAZARO

rlaz133:
Done both.

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -8,7 +8,7 @@
     <!-- don't forget to link your styles -->
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


IH-WebDev-TA-Remote:
Good job but the font sizes could be smaller. 
Also, there seemed to be an issue while accessing the images, try to access them from navigating between the folders and not targeting  the main folder
Do: url('../images/apple-pie.jpg') 
Instead of: url(/lab-css-recipes-clone/apple-pie/images/apple-pie.jpg)
But overall, good job! keep it up.

PR #2138, opened 2020-07-08
WEBDEV - REMOTE - July2020 Joel Roca Martinez

JoelRocaMartinez:
I love html and css


IH-WebDev-TA-Remote:
Very good job overall but be more careful with images sizing and how to center them, you could use flexbox but even with only using margins and padding it is possible.
Also, access images from navigating within the files instead of doing the whole local url.
Do: src="images\apple-pie.jpg"
Instead of: src="C:\Users\Joel Roca\Desktop\Ironhack\lab-css-recipes-clone\apple-pie\images\apple-pie.jpg"

PR #2139, opened 2020-07-08
[WDFT-REM-2007] Elise Dujardin

elisedjn:
Both recipes done. The CSS is almost the same on both pages.


IH-WebDev-TA-Remote:
Looks absolutely perfect! very good job! keep it up! :)

PR #2140, opened 2020-07-08
web-rem-2007 - katie stamey

kmstamey:


IH-WebDev-TA-Remote:
Very good work overall, the page looks very similar!
There is a way to change the list items symbols to dashes and parentheses with some properties. Take a look when we send the solution.
But again, very good job! looks very clean. Keep it up!

PR #2141, opened 2020-07-08
[WEB-REM-2007] GunnerAndersen

GunnerAg:
Only done the mandatory so far


IH-WebDev-TA-Remote:
The images and fonts could be a little bit smaller but overall good job.
Main feedback would be to try to access images by navigating within the files instead of selecting the whole root path.
Do: url('../images/apple-pie.jpg');
Instead of: url(/lab-css-recipes-clone/apple-pie/images/apple-pie.jpg);
Similar for the images in the html.

PR #2142, opened 2020-07-08
WEB-REM-2007 Bodei

BodeiBrouwer:
If you are looking for apple pie or onion soup, this is where to look.
Managed to finish it all and I think it looks pretty decent.


IH-WebDev-TA-Remote:
Very good job! pages looks exactly the same :)
And awesome that you managed to change the symbols in the list items!
Keep it up!

PR #2145, opened 2020-07-08
[WEB-REM-2007] Lanette Lanting

llanting:
Had the feeling that this would've been easier with Flexbox, but after trying it out didn't got it to work for me. So I'll save that for after the lecture of tomorrow. 

Also, I think that I've used way to many classes and selectors. Is this something I have to work on, or is that just how CSS works?


IH-WebDev-TA-Remote:
Very good job! pages look very very similar. Kudos!
If possible you want to avoid unused properties but honestly, don't give it too much though, the important thing right now is that everything looks how it is supposed to.
For the background image I am not sure if it was showing for you but the way to target that would be: 
url('../images/apple-pie.jpg');
Keep up the good work! :)

PR #2146, opened 2020-07-08
[WEB-REM-2007] Marta Lourido

MartaLourido:
hello!

here is my apple pie recipe page, I also did the French soup, pretty much similar.

I really enjoyed working in this lab!

Marta


IH-WebDev-TA-Remote:
Very good job! pages look very similar to the original and great that you managed to change the list item symbols :)
The background image could use some work with additional properties but overall it looks great, keep it up!

PR #2147, opened 2020-07-08
commit css

gonzalo-development:


IH-WebDev-TA-Remote:
Overall looks very good but it could definitely use some improvements.
For start, removing the body border
Also, the recipe text was already in the lab so no ipsum needed!
Finally, responsiveness was not required but still try to work around % sometimes instead of pixels to make it so the images don't look way off when you resize the screen.
Finally, awesome job changing the list item symbols! Keep it up!

PR #2148, opened 2020-07-08
[WDFT-REMOTE-2007] Alejandro Hernández

alejandrohj:
Hi TA,

I think it's close to perfect. I have used the fixed property to put all in the correct way. Maybe the text could be smaller.

Regards,


IH-WebDev-TA-Remote:
Looks really great! definitely very similar to the original, Kudos!
I would just say it is best practice to do urls in quotes, even if css is odd and accept them without, and you don't need to target root elements. So doing ( url("../images/apple-pie.jpg"); ) would work as well.
Keep up the good work!

PR #2157, opened 2020-07-28
[AMS-WEBDEV-JUL27] - Viktor Jensen-Torp

ViktorJT:


french-soup/index.html
```diff
@@ -5,17 +5,51 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
reMRKableDev, line 5: Nice work, however, you've placed this solution in the wrong folder. It should be in ```apple-pie``` and not ```french-soup```.

french-soup/styles/style.css
```diff
@@ -7,3 +7,44 @@ html {
   font-family: 'Open Sans', sans-serif;
```
reMRKableDev, line 7: Same comment here... nice work, however, you've placed this solution in the wrong folder. It should be in ```apple-pie``` and not ```french-soup```.

PR #2158, opened 2020-07-29
[AMS - WEBDEV - 072020] - Esra OZEL

Esra90:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,63 @@
...
+#directions {
+    display: block;
+    height: 300px;
+    width: 1000px;
+    /* border: 2px solid black; */
+    margin:0 auto;
+    padding-top:15px;
+}
+
+#
```
reMRKableDev, line 63: Not sure if this ```#``` is necessary/needed here.

PR #2202, opened 2020-08-25
[PAR-WDFT-248] FRANCOIS VIEUX

Georges-Seonho:


Fanny-Sbn:
Hello François, thank you for submitting the first lab ! It looks great, just a few comments. 
You need to remove the margin and padding of the page that the browser you use automatically applies. 

*{
margin: 0;
padding:0;
}

To center the title with the image background you can wrap your title in a div, then apply the background-image property (and the others) on this div and use flexbox to center the title. 

You’re using the width property with the vh value (width: 95vh;) – vh means view height and vw means view width. I think that the one you wanted to use. 

You maybe not pay attention, for the first list we want a dash before each step of the recipe and not a dot and for the second one we want a parenthesis and not a dot. Try to find on google how to change the list style 😉

Good job 👍

Fanny-Sbn:
Hello François, thanks for submitting a new version of the lab. A few comments 😄 

You haven't modified the apple pie version... but nice for doing the second one ! 👌 

You changed the vh as I told you and found the way to put the dashes for the first list : super great ! 
But for the second list, you also put dashes instead of a number and parentheses. 

Good job anyway !

PR #2203, opened 2020-08-26
[PAR-WDFT-248] LAURA MANTAL-D'ABOVILLE

Lauradabo:


Fanny-Sbn:
Hello Laura, thank you for submitting the first lab 😊 It looks super great!

The only difference with the original website is the first paragraph which is supposed to be align with the ingredients and the descriptions, and take more space on the width of the page. The rest is perfect!

Your code is well structured, nothing specific to say about that.

Just a few more things:

•	I saw that you moved your css file and put it at the same level of your index.html. In a near future, you’ll have more than one css file so storing it in a styles folder is a good practice. To access it from your html file you just need to change the href in the link tag which linked your files. So move your css file into the styles folder and change the href of your link tag in the head for href="./styles/style.css".

•	You need to remove the margin and padding of the page that the browser you use automatically applies in your css file. 
*{
margin: 0;
padding:0;
}

•	Finally, you avoid the “tricky” part of the lab writing directly the dashes and the numbers with parentheses in front of the elements of the lists in the html file. You should have found a way to do it in CSS. Try to find the solution by yourself but if you don’t, Guillaume will give you the correction.  

Super good job Laura ! 👍 💃

PR #2204, opened 2020-08-30
[PAR-WDFT-248] Laura Caillaux

Lcdevi:


Fanny-Sbn:
Hello Laura, thanks for submitting the first lab. 

I don't have a lot to say, it looks super good ! 

In your CSS file, instead of adding a padding to your div #recipe on left and right sides, you can set a width. 
#recipe {
    width: 90vw;
    padding: 90px;
}

Just one thing, you avoid the “tricky” part of the lab writing directly the dashes and the numbers with parentheses in front of the elements of the lists in the html file. You should have found a way to do it in CSS. Try to find the solution by yourself but if you don’t, Guillaume will give you the correction.

Nice job ! 💯

PR #2215, opened 2020-09-16
[REM-2009]zuzana 

ZuzanaAK:


IH-WebDev-TA-Remote:
Very good job! The clone is very similar to the original and there is some responsiveness to it. Also correct use of tags and well indented. Keep it up!

PR #2216, opened 2020-09-16
[REM-2009] Sandra Mendez

scmendez:
The LAB felt fairly straight forward, not sure if I did the Directions list correct though. I couldn't find how to change from 1. -> 1), so I added it to the text instead.

Also: I understood "Extra Practice" to mean optional, my mistake. Added the French Onion Soup part just now though.


IH-WebDev-TA-Remote:
Really good job! the clone looks like the original and there is good responsiveness. Also good use of tags and classes. 
The way you added the "1)" is actually good. However, there is a way to add the symbol directly through css. Check the solution we send later for that.
Still very good, job, also the code is well indented and spaced. Keep it up!

PR #2217, opened 2020-09-16
[REM-2009] Aleksandra Tanasijenko

alekstana:
:)


IH-WebDev-TA-Remote:
Good job! the page look very good.  Good use of tags and classes. I see you had some issues with the background image but that is usually one of the harder parts haha. Check the solution we will send later to see how to center it correctly with the right segment showing.
Also, there is a way to add the "1)" to the list items with css, check the solution for that as well.
Overall though, well done! Keep it up!

PR #2218, opened 2020-09-16
[REM-2009] Sumiya Ushiro

sumi0820:
I completed Apple pies and French onion soup.

I struggled with background images and customized lists. I did not know there is a specific attribute to set images as background easily, so it is good to know. As for the lists, I also did not know that we can set " - " or customized property. It was a good LAB to learn new things.

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -3,19 +3,78 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="./styles/style.css" />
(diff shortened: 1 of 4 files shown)


IH-WebDev-TA-Remote:
Good job, the page looks fine in full desktop. And I see you managed to change the list symbols via css so kudos! 

Now, the background image is a little big and It is not very responsive but that was not a requirement so don't worry about it. For practicing responsiveness and even media queries we will send another bonus Lab later.

The background image and the list symbols are easily the hardest parts so struggling on that is normal, also yes, a little bit of research was needed ;)

Overall though, well done. Keep it up!

PR #2219, opened 2020-09-16
[REM-2009] Laura Salakari

LauraSalakari:
CSS is hard T.T


IH-WebDev-TA-Remote:
Very good job! Page looks very like the original one! And there is even some responsiveness to it. CSS is definitely hard! Something you will hear even from pro devs xD

Good work on centering the background image. Good use of tags, and the code is well indented and spaced. How you did the list item symbols is perfectly fine but there is a way to add the "1)" via css, take a look at the solution we will send to see how!

Overall though very good job, keep it up!

PR #2220, opened 2020-09-16
[REM 2009] Ross Homaidan

ibirossi:
I had some problems with silly mistakes, but not too bad.  Definitely need to practice more over weekend, to get things fixed in my mind.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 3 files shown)


IH-WebDev-TA-Remote:
Good job! the page looks somehow like the original. Very well done on the background image, looks great! definitely keep practicing html&css is encouraged! not only with this lab but also with the spotify one and two more bonus ones we will send later.
Main feedback would be regarding code cleanness for the html. Watch spacing and be more careful with indentation of elements, moving them one tab to the side when they are inside another tag. See the solution for how it should be structured.
Overall though, good job. Keep it up!

PR #2221, opened 2020-09-16
[rem 2009] katie stamey

kmstamey:


IH-WebDev-TA-Remote:
Good job! The background image and title box could have been better but at least in one screen size it works ok. A better approach to have it centered and more responsive would be to work around percentages and flexbox and not have hard coded sizes. Yet, everything below that looks very good and is very close to the original!

Another feedback would be to improve the html code cleanness. Have better spacing and indentation of items inside other tags. Check out to solution we will send for this and also to see how the title box is handled and even how to add the "1)" to the list items via css.

Keep practicing those html&css skills!

PR #2222, opened 2020-09-16
[REM - 2009]Marta Obregon

MartaObregon:


IH-WebDev-TA-Remote:
Great work!! the page looks very close to the original and even with some responsiveness. Code is well indented and even commented (which is always appreciated!). Tag use and class naming is good.

Putting the list item symbols manually is ok, however there is a way to add them via css which is usually preferred. Take a look at the solution once we send it to see how.

Again, amazing work! Keep it up!

PR #2223, opened 2020-09-16
[REM - 2009] Vykintas Petrikas

VykiPetr:
I think i did both of the pages!

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -1,21 +1,60 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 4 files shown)


IH-WebDev-TA-Remote:
Good work! The page looks very similar to the original one. It could use a little more polish on the responsiveness side but it was not a requirement for this lab. Still I would suggest you keep practicing with using percentages and flexbox to achieve that, on this lab or even better on the spotify one and a couple more bonuses we will share later.

Finally, similar to adding the "- " to the list items there is a way to add the "1)" via css. Check the solution we will send later to see how. And for the background image url, it is better to access the image from within the files like this ( background-image: url("../images/apple-pie.jpg"); ). This is to make it work if someone else clones your repo or even when deployed.

Still, very well done. Keep up the great work!

PR #2224, opened 2020-09-16
[REM-2009] Jordane Ruiz

jordaneruiz:
Apple Pie Lab done


IH-WebDev-TA-Remote:
Good effort on the lab but still the page doesn't look quite similar to the original. There seem to be some code there in html and css that is not being used. Also, there are some things disproportioned in the page. 

I would really suggest you go through the solution we will share and see how the css attributes are set. Try to change the sizes to make it better proportionate and work around flexbox instead of full block displays.

PR #2225, opened 2020-09-16
[REM 2009] Lovro Hodak

LovroHodak:
bum-chicka-wuawua


IH-WebDev-TA-Remote:
bum-chicka-wuawua indeed. Good job! the clone looks very similar to the original.  The code well indented and easy to read.

Very good job changing the symbol of the first list! There is also a way to add the "1)" to the second one via css. check out the solution to see how.

But overall very well done. Keep it up!

PR #2226, opened 2020-09-16
[REM-2009] Santiago Andrés Ardiles Páez

SantAndresP:
I didn't find a way to change the default **_<ol>_** formatting of **_1._** to **_1)_**.


IH-WebDev-TA-Remote:
Amazing job! The page is the same as the original and even there is some responsiveness to it.
I only see you switched the images but that is ok hahaha. Perfect indentation and spacing, makes the html very easy to read.

There is actually a way to do the "1)" part of the list with css, check the solution we will send later to see how. 

Again, amazingly done. Keep it up!

PR #2372, opened 2020-10-21
[WDFTRM 20201021] - Monique

moniquedehaard:
It's not really beautiful, i am a little bit struggling with padding, margin, and than using flexbox (and why images will not center..). Also i'm a little bit struggling with vw, vh, % or using px.. What is the best way for using center?

### DIFF
--- a/apple-pie/styles/style.css
+++ b/apple-pie/styles/style.css
@@ -1,0 +1,73 @@
+@import url("https://fonts.googleapis.com/css?family=Open+Sans");
+
+* {
+    padding: 0;
+    margin: 0;
+    box-sizing: border-box;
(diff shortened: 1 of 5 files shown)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,79 @@
...
+    justify-content: space-between; 
+}
+
+.text {
+    font-size: 25px;
+}
+
+.recipe-img{
+    height: 60px;
+    width: 300px;
```
filipe-freire, line 44: try using `margin: 0 auto;` here and see the outcome!

filipe-freire (approved):

PR #2431, opened 2020-11-09
[PA WDFT 911 Calypso Asline]

Calyaln:
pbm:
- resizing all 3 images
- not using flex or grid is a challenge...


Calyaln:
restarted from scratch after reflecting and it's much better!

PR #2446, opened 2020-11-11
WDFT911 - Intan PAOLI - Apple pie recipe clone 

np94:
Hi TAs, 

I did not manage to import the fonts needed. Maybe the picture was too big even when i have applied vw/vh. I know its not perfect, but i will try better next one.

Cheers, 
Intan

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -9,8 +9,8 @@
 
   <header>
     <div class="header">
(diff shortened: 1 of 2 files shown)


np94:
Hi, 

I added the hr, the apple pie word on the photo using z-index and added the font. only the second divider is thicker than the first one. 

cheers, 
Intan

PR #2690, opened 2021-06-02
Liliia - WDFT - RMT - MAY21 - DAY3

liliiakhr:


TA-Remote:
Nice job on the lab. For the background image have a look at this link to see how you could have done it with CSS only : https://www.w3schools.com/cssref/pr_background-position.asp

PR #2691, opened 2021-06-02
RITA MAK WDFT - RTM - MAY21 (lab-css-recipes-clone)

ritamak:


TA-Remote:
Good job Rita! Its pretty close to the original! Keep practicing a little bit flex bot and try to understand where, when and how to use it. The header image should also be smaller, always try to be as exact as possible to the original. Very good job anyways! Keep the hard work! 💯 🏅

PR #2692, opened 2021-06-02
WDFT-AMS-2021 AnnaSun

AnnaSun919:


TA-Remote:
Good job Anna! Very close to the original exercise!

PR #2698, opened 2021-06-02
[WDFT-RMT-MAY21] Conchita  

CodeChita:
Hi TA's!

CSS is a pain in my **s lol! 
I made it but don't ask me to do it without any google privileges. 
Hope you have a nice day :)

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -1,21 +1,47 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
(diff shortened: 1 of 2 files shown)


TA-Remote:
Hi Conchita! CSS can be a pain in the **s! With practice it get easier and easier.  Good job Conchita! Really close to the original! You should have tried to center everything to the middle and give it a max-width so it doenst look so wide. Anyway! Almost perfect! 💯

PR #2701, opened 2021-06-02
KALANDE-WDFT-RMT-MAY2021

Kalande:
Apologies for the delay, had a problem with windows that I had to resolve. : )


TA-Remote:
Nice lab. For the images it was just a small mistake of url path. Check the solution later to see how you could have written the path to show the images and feel free to ask question if you have any doubt :)

PR #2702, opened 2021-06-03
Website "clone"

r00kie123:
SInah Scholz WDFT-RMT-MAY21


TA-Remote:
Nice lab. Feel free to check this link and the solution later to see how you could have made the background image bigger https://www.w3schools.com/cssref/pr_background-position.asp

PR #2913, opened 2022-01-12
Marian Gispert WDFT-JAN 2022

mgispert:


TA-Remote:
Also extremely good! It doesn't look 100% like the picture, but very very close and what's much more valuable: it's responsive and looks great overall!

PR #2997, opened 2022-04-13
[FTWD042022] Catarina Salvado

catarinapsalvado:


webLisbonTA (commented):
Good job on the clone!
A quick note:
- The image on top should be a cover with 100% width

PR #3001, opened 2022-04-14
[FTLIS042022] Chloé Faurie

chloe4E:
- Apple pie done
- French soup almost done.

Issue: I would like to use only one css for both recipe. My remaining challenge is the background image. I am struggling to change the logic not to have it in css but in html to be able to use the same css for both recipes.
@[webLisbonTA](https://github.com/webLisbonTA) do you have an idea ?

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -5,47 +5,56 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Apple Pie Recipe</title>
     <!-- don't forget to link your styles -->
-    <link rel="stylesheet" href="styles/style.css">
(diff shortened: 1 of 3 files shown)


webLisbonTA (commented):
Great job on doing both exercises!
Some quick notes:
- When writing for websites be careful with the font, sometimes it can be hard to read due to its size;
- The text should be centered on the first paragraph;
- Regarding the images, the instructions and preparations should be a bit smaller;

On the french soup it's giving you an error because you have the apple image on your CSS file and the french soup on an html. To fix that you should use both images in html so it doesn't have any issues. 

Nonetheless, you did a very good job!

PR #3003, opened 2022-04-14
[FTLIS042022] Alexandre Alves

aletrad:


webLisbonTA (commented):
Good job on doing the clone! For future exercises try and look at "rem, em and %" for sizing measures, it's better practice than using px. Nonetheless, good job!

PR #3007, opened 2022-04-14
[FTLIS042022] matildeCosta

matildedacosta:


webLisbonTA (commented):
Great job on finishing both exercises!
You didn't use the best practice for linking images, Andre will talk about it in class!

PR #3008, opened 2022-04-14
[wd_ft_apr_2022]Mariana Pontes

mariannasp8:


webLisbonTA (commented):
Good job on the clone!
Be careful with the font you choose, some fonts are not very readable due to its weight and color.
The images were't linked properly so they were't showing up.

PR #3010, opened 2022-04-14
[FTLIS042022] anaPinho

AnaGonPinho:


webLisbonTA (commented):
Good job on finishing the clone!
Some quick notes:
- The image should be a bit bigger as we saw on the lab instructions;
- The first paragraph should be centered
- The preparation image should be smaller

Nonetheless, you did a good job with the clone using great selectors in css

PR #3068, opened 2022-06-13
[RMFT JUN2022] Ruben Andino

rubenanlo:
Thanks for reviewing it!! Any improvement will be much appreciated.

### DIFF
--- a/apple-pie/styles/style.css
+++ b/apple-pie/styles/style.css
@@ -21,6 +21,8 @@ section {
     border-bottom: 1px solid grey;
     padding: 1.5rem 3.5rem 1.5rem 3.5rem;
     margin: 0 3.5rem 0 3.5rem;
+    animation-name: backInLeft;
+    animation-duration: 2s;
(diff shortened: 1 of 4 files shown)


TA-Remote:
Impressive! really good job!
just a minor thing about image path :)

apple-pie/styles/style.css
```diff
@@ -0,0 +1,84 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@300&display=swap');
+
+body {
+    font-family: Roboto;
+    margin: 0;
+    padding: 0;
+}
+
+header {
+    background-image: url("/apple-pie/images/apple-pie.jpg");
```
TA-Remote, line 10: When you write path here, remember that you need to go to the image from the file where you define the path , in this case from style.css file.
 style.css ->styles -> images -> apple-pie.jpg

 background-image: url("../images/apple-pie.jpg")

PR #3070, opened 2022-06-13
Alexandre Coma pull request

Virgule09:
Alexandre Coma


apple-pie/styles/style.css
```diff
@@ -1,6 +1,7 @@
 body {
     margin: 0px;
     width: 600px;
+    font-family: "Proxima Nova Light";
 }
```
TA-Remote, line 5: A suggestion here, if you write 
`margin: 0px auto;` or `margin: auto;`

This would align the content of the body 'automatically' in the center of the browser window. In that case, you could increase width . For regular screen it will be full width and for the very wide screen - it will only fill the middle of the window

apple-pie/index.html
```diff
@@ -4,18 +4,56 @@
...
-    C). Melt the butter in a saucepan. Stir in flour to form a paste. Add water, white sugar and brown sugar, and bring
-    to a boil. Reduce temperature and let simmer. Place the bottom crust in your pan. Fill with apples, mounded
-    slightly. Cover with a lattice work crust. Gently pour the sugar and butter liquid over the crust. Pour slowly so
-    that it does not run off. Bake 15 minutes in the preheated oven. Reduce the temperature to 350 degrees F (175
-    degrees C). Continue baking for 35 to 45 minutes, until apples are soft.
+   
+    <section id="apple-pie-image">
+      <img src="./images/apple-pie.jpg" alt="apple pie image"/> 
+      <h1>Apple Pie</h1> 
+    </section>
```
TA-Remote, line 14: - With the html structure that you have, you should use positioning to align elements. You need to set position `relative` to parent element and position `absolute` to child element . By default child element will sit at the top left corner of the parent element 

Example below:

```
.parent {
  position: relative;
}
.child {
  position: absolute;
  top: 50%;
  left: 50%;
  transform: translate(-50%, -50%);
}
```

- Another approach could be to set background image for your section and only align the h1 inside it.

`background-image: url("../images/apple-pie.jpg");`

TA-Remote (commented):
Great work Alex!  Gave couple sugesstions how to improve - how center the title inside the image, if you try and still does not work , let me know 
Karina :)

PR #3072, opened 2022-06-13
[RMFT JUN2022] Paul Tiry

ptiry02:
Hi, I just wanted to let you know that I already have a few months of experience with frontend developing. That is why I finished the assignment so early. 😅


TA-Remote:
Great! really god job!
so you can make this Lab responsive or do french soup lab if you want ,, 
I mean you can always work on Extra/bonus things with Labs if you want  ;)

PR #3074, opened 2022-06-13
Lara Rockenstein

LMRBLN:


LMRBLN:
**Open questions**
Why in my ordered list, the text is surpassing the borders of the parent box?
Why does nth-of-type not work for img (see what I wrote as a comment in the css-file)?

PR #3077, opened 2022-06-13
Tobias Conrad

maktooob:
Good morning! Hope that works :)


TA-Remote:
Well done Tobias,

apple-pie/styles/style.css
```diff
@@ -0,0 +1,61 @@
+body {
+    margin: 1rem;
+    padding: 0;
+    font-family:'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+    font-size: 12px;
+    color: rgb(96, 96, 96);
+    max-width: 500px;
```
TA-Remote, line 7: we want to see the data in the whole page , so we dont need to define width or max-width here

apple-pie/styles/style.css
```diff
@@ -0,0 +1,61 @@
...
+    line-height: 1.1rem;
+    
+}
+main {
```
TA-Remote, line 11: after removing max-width, you need to change margin numbers to get good result ;)

PR #3081, opened 2022-06-13
Julie Bardiani "apple-pie, done"

Nika-Julie:


Nika-Julie:
- I found no way to put the border-bottom & border-bottom-color. didnt take it in no sections or extra created divs
- The margin from the first h2 floated into the picture so no distance was created. Had to put a margin-bottom to the picture to gain the gap and i dont know why  
Thank you for review :)

PR #3082, opened 2022-06-13
Stephanie Tovar - first lab on HTML/CSS completed

Stephanie-Tovar:


TA-Remote:
Really great Job !
just on thing :
when we want to use * selector, because it apply the style to all elements, we write it normally at the top of  the CSS document

PR #3083, opened 2022-06-13
[RMFT June '22] Lisa Hilterhaus: implemented recipe page apple pie

cortinaHs:


TA-Remote (commented):
Great job! Your code looks really clean and well organized. The site is responsive too.

Just one pointer, if you change the font-size we lose the vertical alignment, but after Day 2 lessons we could solve this problem with other solutions :)

PR #3114, opened 2022-07-04
[FTLIS072022] John Imeson

johnimeson0:
.

### DIFF
--- a/apple-pie/index.html
+++ b/apple-pie/index.html
@@ -3,19 +3,61 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+   <link rel="stylesheet" href="styles/style.css">
     <title>Apple Pie Recipe</title>
     <!-- don't forget to link your styles -->
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


apple-pie/index.html
```diff
@@ -20,6 +21,7 @@
     favorites as well! <br />
 
     </div>  
+      </header>
     <div class="recipe-info">
       <img src="/Users/johnimeson/lab-css-recipes-clone/apple-pie/images/recipe-info.png"
```
webLisbonTA, line 26: When we are using images locally we specify the path in a different way:
```
<img src="images/recipe-info.png" />
```

apple-pie/styles/style.css
```diff
@@ -14,16 +14,25 @@ h1, body, html, main {
...
 
   
-  .top-of-page {
+   .top-of-page {
     width: 1440px;
     height: 825px; 
     font-size:100px;
     font-weight:lighter;
     color:white;
     background-image: url(/Users/johnimeson/lab-css-recipes-clone/apple-pie/images/apple-pie.jpg);
```
webLisbonTA, line 23: When we are using images locally we specify the path in a different way:
```
background-image: url("../images/recipe-info.png")
```

PR #3118, opened 2022-07-04
[FTLIS072022] Adriana Leitão

AdrianaLeitao:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,50 @@
+body {
+    margin: 50px;
```
webLisbonTA, line 2: Here, you could give the body a margin-left and margin-right of 100px instead of repeating it in the other classes, like this:
```
body {
margin-left: 100px;
margin-right: 100px;
}
```

webLisbonTA (commented):
Great job on the lab, everything was according to the clone!

PR #3119, opened 2022-07-04
[FTLIS072022] Simão Lima

Simaodosreislima:


webLisbonTA (commented):
Good job on the lab! Regarding the placement, it's appearing like that because on multiple classes you gave different margins to certain divs / elements meaning that the alignment would be different for each element. Good job nonetheless!

PR #3121, opened 2022-07-04
[FITLIS072022] Majid Boukachni

magicmajid2511:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,29 @@
+.applepie {
+    background-image: url("../images/apple-pie.jpg");
+    background-size: cover;
+    background-position: center;
+    width: 1000px;
+    height: 753px;
+}
+.recipe {
+    width: 1000px;
+    height: 753px;
```
webLisbonTA, line 10: Here you didn't need to add a height since you already added on the class before

webLisbonTA (commented):
Great job on the lab, everything was according to the clone!

PR #3123, opened 2022-07-04
[FTLIS072022] Luísa Coelho

marilu777:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,14 @@
+.title{background-image:url("../images/apple-pie.jpg");
+    background-size: cover;
+    height: 500px;
```
webLisbonTA, line 3: Here you could replace the height with a padding of 250px and that would center the image automatically

webLisbonTA (commented):
Great job on the lab, everything was good!

PR #3144, opened 2022-07-14
done

JoaoSerafim3001:


apple-pie/index.html
```diff
@@ -5,17 +5,71 @@
...
+          <hr>
+        </div>
+      </article>
+      
+      <article>
+        <div>
+          <h3>Ingredients</h3>
+
+          <ul>
+            <li>- 1 recipe pastry for a 9 inch double crust pie</li>
```
trebeljahr, line 40: if you do it this way with the minus, add a list-style: none; to the ul element.

apple-pie/index.html
```diff
@@ -5,17 +5,71 @@
...
+        </div>  
+     
+      </article>
+
+      <article>
+        <div>
+          <h3>Directions</h3>
+
+          <ol>
+            <li>1) Preheat oven to 425 degrees F (220 degrees C). Melt the butter in a saucepan. Stir in flour to form a paste. Add water, white sugar and brown sugar, and bring to a boil. Reduce temperature and let simmer.</li>
```
trebeljahr, line 65: same as with the ul comment above, list-style: none; should get rid of the duplicate list styling.
furthermore you might want to have another look at how we did it today during lab review. Because your approach might break easily and is tedious to keep up, imagine if your recipe had 100 steps ^^ what about a 1000? or 100000?

apple-pie/index.html
```diff
@@ -5,17 +5,71 @@
...
+
+    </section>
+    
+    <section id="main-content">
+      <article>
+        <div>
+          <h4>This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always be my favorite and has won me several first place prizes in local competitions. I hope it becomes one of your favorites as well!</h3>
+          
+          <div class="img-container">
+            <img class="info-img" src="/lab-css-recipes-clone/apple-pie/images/recipe-info.png"/>
```
trebeljahr, line 28: src for the images is wrong -> should be only "images/recipe-info.png"

apple-pie/styles/style.css
```diff
@@ -0,0 +1,54 @@
...
+    font-family: 'Open Sans', sans-serif;
+    font-size: 16px;
+}
+
+ul, ol {
+    list-style: none;
+}
+
+#header {
+  background-image: url(/lab-css-recipes-clone/apple-pie/images/apple-pie.jpg);
```
trebeljahr, line 17: this one is even more trickier, when it comes to picking the correct url because we are already within /styles. 
Best to read this again:
https://stackoverflow.com/questions/2005079/absolute-vs-relative-urls
and think about what that means in this case. 
Right would be "/apple-pie/images/apple-pie.jpg"

trebeljahr (commented):
Nice work :) :+1:

apple-pie/index.html
JoaoSerafim3001: Yes, I searched for the code but since we didn't mentioned at the class I cheated a little bit☺.

apple-pie/index.html
JoaoSerafim3001: The same here..

apple-pie/index.html
JoaoSerafim3001: Thank you

apple-pie/styles/style.css
JoaoSerafim3001: Thank you.

PR #3145, opened 2022-07-15
Project finished with other background image

herrFernandez:
I finished the project including an image from the internet for the apple pie background. Somehow I couldn't manage to display the image from the path.

The path I used: /apple-pie/images/apple-pie.jpg following visual studio path


trebeljahr (commented):
Really good work :) :+1: keep it up like this!

PR #3146, opened 2022-07-15
WDPTJul22 Patrick

Patrick-Groot:


apple-pie/index.html
```diff
@@ -4,18 +4,54 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Apple Pie Recipe</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="styles/style.css">
   </head>
+  <header>
+    <img src="/home/patrick/Schreibtisch/IronHack/lab-css-recipes-clone/apple-pie/images/apple-pie.jpg" alt="apple pie" id="background">
```
trebeljahr, line 10: src references like this will break on every other machine! 
It's relative to your exact file system and nobody else's will look like this. 
Try using relative imports instead. src="images/apple-pie.jpg"

apple-pie/styles/style.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+ol {
+    list-style-type: none;
+    counter-reset: list;
+}
+ol > li:before {
+    counter-increment: list;
+    content: counter(list, decimal) ") ";
+}
```
trebeljahr, line 68: good work :+1:

french-soup/index.html
```diff
@@ -4,18 +4,50 @@
...
-    thyme. Season with salt and pepper, and simmer for 30 minutes. Heat the oven broiler. Ladle soup into oven safe
-    serving bowls and place one slice of bread on top of each (bread may be broken into pieces if you prefer). Layer
-    each slice of bread with a slice of provolone, 1/2 slice diced Swiss and 1 tablespoon Parmesan cheese. Place bowls
-    on cookie sheet and broil in the preheated oven until cheese bubbles and browns slightly.
+    <section id="intro">
+      <article>
+        We have been trying French onion soup in restaurants for years and my family and friends agree
+        none can compare to my recipe for taste and simplicity of preparation.
+      </article>
+      <img src="./images/recipe-info.png" alt="recipe-info">
```
trebeljahr, line 19: this is much better than the above. try getting rid of the "./" as well and it would be even better. 
Reason: Not in every filesystem you have something like the "." and ".." directories, some servers might not implement those shortcuts for security reasons, mainly as a measure against directory traversal attacks!

french-soup/styles/style.css
```diff
@@ -7,3 +7,62 @@ html {
...
+  align-self: center;
+}
+
+#directions {
+  margin: 0 10vw 0 10vw;
+}
+hr {
+  height: 1px;
+  color: black;
+  width: 100vw;
```
trebeljahr, line 54: this should be 100% instead I think. That way it would size according to it's parent width, for me with 100vw it actually wraps off the screen because of the left and right margins of the parent!

trebeljahr (commented):
except those minor things in the comments, this is really good work! even with the bonus iteration of the french soup - looks good, keep it up :)

french-soup/styles/style.css
Patrick-Groot: Thank you for the helpful feedback!

PR #3147, opened 2022-07-15
 [PTBER072022] berndHavekost

hvkst:
everything done, added a little responsiveness


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "liveServer.settings.port": 5501
```
trebeljahr, line 2: this doesn't need to be committed ^^

apple-pie/index.html
```diff
@@ -4,18 +4,79 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Apple Pie Recipe</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="styles/style.css" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
```
trebeljahr, line 10: nice font import :)

apple-pie/index.html
```diff
@@ -4,18 +4,79 @@
...
+            <li>1/2 cup white sugar</li>
+            <li>1/2 cup packed brown sugar</li>
+            <li>8 Granny Smith apples - peeled, cored and sliced</li>
+          </ul>
+        </section>
+
+        <section class="howto">
+          <img
+            id="recipe-info"
+            src="/apple-pie/images/cooking-info.png"
```
trebeljahr, line 59: depending on which part of a page (which URL you are on) the / before file srcs might not be what you want. 
have a look at this: 
https://webmasters.stackexchange.com/questions/56840/what-is-the-purpose-of-leading-slash-in-html-urls
tldr: it might break when you are nested in sites and your image src is not nested :+1:

apple-pie/styles/style.css
```diff
@@ -0,0 +1,147 @@
+body,
+html {
+  margin: 0;
+  padding: 0;
+  /* font-family: "Roboto", sans-serif; */
+  font-family: "Open Sans", sans-serif;
```
trebeljahr, line 6: use the family that you imported from google :)

trebeljahr (commented):
really good job :) outstanding work. keep it up like this :+1:

apple-pie/styles/style.css
hvkst: Yeah, I wasn´t really happy with Roboto and in the French Soup css was Open Sans, so I used that in both :)

PR #3210, opened 2022-09-05
[RMT -5SEP] Saskia Peach

saskiap97:


saskiap97:
Following Felipe's tutorial in the lab

saskiap97:
One thing I am confused about is that I wanted to make the images between the lists smaller, but if I used the width: property in CSS it would move my image on the page! I don't know how to make the images smaller without affecting their placement on the page.

PR #3214, opened 2022-09-05
[RMT-5SEP]FARZIN-KHALEDI

FarzinKhaledi:


TA-Remote (commented):
As Luis said in the beggining of the class the images were inside the folder.

Aside from that try work a little bit more on position of the elements

PR #3216, opened 2022-09-05
[RMT 5 SEP] Matthew Schröder 

MatthewS9407:
I couldn't get the font as thin as in the example.
I also had some trouble with git push origin master, but I think i figured it out.
Hope it worked :)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,45 @@
+body {
+    font-family: Arial;
+    font-size: 1em;
+    font-weight: lighter;
+    
+}
+
+.header {
+    
+    background-image: url("/apple-pie/images/apple-pie.jpg");
```
TA-Remote, line 10: ` background-image: url("../images/apple-pie.jpg");` that way the image would be displayed

TA-Remote (commented):
Great work!

Don't mind getting 100% perfection on the clones :)

PR #3220, opened 2022-09-05
[RMT 5Sept]_Daniel Plesa

DanielPlesa:
Its late and I couldn't figure out how to align the header text over image. The rest I'm quite happy with (i think).

Thanks for any feedback!


TA-Remote (commented):
Great work, It could be added some styles for the fonts and trying some spacing.

- For the alignment you could have used flexbox. 
- Or a `left: 40%` or another number, because you already have the `relative `and `absolute`

PR #3222, opened 2022-09-05
[RMT -5SEP] Julia Spriggs

julia-spriggs:
For some reason I can't seem to get the background image to show up for h1, even though I know I did it successfully in the past. I've checked repeatedly to make sure my path was correct, and it seems to be (for example, I put the exact same pathj in my HTML to test and the image worked there). So I'm not sure what I did wrong.

### DIFF
--- a/apple-pie/styles/style.css
+++ b/apple-pie/styles/style.css
@@ -1,0 +1,35 @@
+body {
+    margin: 1em 10%;
+    font-family: Arial;
+}
+
+h1 {
+    background-image: url("images/apple-pie.jpg");
+    position: 0 0;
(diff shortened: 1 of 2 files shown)


apple-pie/styles/style.css
```diff
@@ -0,0 +1,35 @@
+body {
+    margin: 1em 10%;
+    font-family: Arial;
+}
+
+h1 {
+    background-image: url("images/apple-pie.jpg");
```
TA-Remote, line 7: You missed some ../ => `background-image: url("../images/apple-pie.jpg");`

PR #3223, opened 2022-09-05
[RMT-FT-092022]-Berkay-Yildirim

berkayyildirim:
Lab1-HTML and CSS Assignment.


TA-Remote (commented):
Very good work.

Try to use less IDs as CSS Selectors, if you need to change something the specificity could make it harder to get what you want to do

PR #3224, opened 2022-09-06
[RMT -5SEP]Haridha

HaridhaRavi:
-Couldn't not match with the exact font style
-please comment on alternative ways to optimize the code


TA-Remote (commented):
Actually it looks really clean and good ^^

PR #3227, opened 2022-09-06
[RMT-5 SEPT] Liz Manresa

9liz9:


TA-Remote (commented):
Good work! ^^ 

- Remember to indent yout code, so it's easier to read.
- Try to use most classes or elements as selectors for CSS, ids can throw some unexpected problems because of the specificity

PR #3237, opened 2022-09-10
[lab-css-recipes-clone] José Alexandre Pereira

4kiUta:


4kiUta:
I didn't use background image to make the title of the recipe in the middle of the element. I decided to try to place an element above another element.

PR #3527, opened 2023-01-18
[LISFT0123] Miguel Jesus

Jizzuchrist:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,45 @@
...
+}
+
+section{
+    margin: 40px 120px;
+}
+
+.cooking-img{
+    text-align: center;
+    border-bottom-style: solid;
+    border-width: 2px;
```
webLisbonTA, line 27: Instead of the border bottom, you should've use an `</hr>` tag in the HTML.

webLisbonTA (commented):
Great job on the lab!

PR #3529, opened 2023-01-18
[LISFT0123] + Assebe Kleinert

Assebe:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,62 @@
+* {
+
+font-family:sans-serif;
+font-weight: 200;
+margin-bottom: 20px;
```
joaopdg, line 5: To get rid of the white border around the image you just need to add this line `margin: 0` here at the universal selecter `*`:

```
* {
  font-family: sans-serif;
  font-weight: 200;
  margin-bottom: 20px;
  margin: 0
}
```

joaopdg (commented):
Great job on the lab Assebe!

PR #3530, opened 2023-01-18
[LISFT0123] Déborah 

deborahdelimaa:
lab done


apple-pie/index.html
```diff
@@ -4,18 +4,70 @@
...
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="/apple-pie/styles/style.css">
   </head>
   <body>
-    Apple Pies This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always
+    <div class="background-image">
+<h1 class="title">Apple Pie</h1>
+    </div>
+    <div>
+<section>
```
joaopdg, line 14: A `<div>` or a `<section>` do the exact same thing, it's just a matter of semantics/good practices, and it is important for people that use screen readers. 
So there is no need to open a `<div>` tag every time you have a `<section>` you can just use the `<section>`:

This:
```
    <div>
      <section>
        <p>
          This was my grandmother's apple pie recipe. I have never seen another
          one <br />
          quite like it. It will always be my favorite and has won me several
          first place <br />
          prizes in local competitions. I hope it becomes one of your favorites
          as well!
        </p>
      </section>
    </div>
```

Is the same as this:
```
      <section>
        <p>
          This was my grandmother's apple pie recipe. I have never seen another
          one <br />
          quite like it. It will always be my favorite and has won me several
          first place <br />
          prizes in local competitions. I hope it becomes one of your favorites
          as well!
        </p>
      </section>
```

apple-pie/styles/style.css
```diff
@@ -0,0 +1,84 @@
+* {
+    margin-top: 0%;
```
joaopdg, line 2: To get rid of the white bars around the background image instead of `margin-top: 0%;` that will only take out the top margin, you can do `margin: 0px` and in this way you will take out all margins (top, right, bottom and left).

apple-pie/styles/style.css
```diff
@@ -0,0 +1,84 @@
...
+} 
+
+.organized-list {
+    list-style-type: none;
+}
+
+.organized-list li {
+    font-size: 18px;
+    width: 50%;
+}
```
joaopdg, line 84: You can add some margins at the left so your text will be a little more centered. So at the end of the file you can add this:

```
hr{
    margin-left: 100px;
}

h3, ol, ul{
    margin-left: 150px;
}
```

joaopdg (commented):
Great job on the lab Déborah!

PR #3535, opened 2023-01-18
finished version

evelinecoen:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,68 @@
...
+	box-sizing: border-box;
+	margin: 0 10%;
+	padding: 0;
+    font-family: proxima-nova, sans-serif;
+    font-style: normal;
+    font-weight: 100;
+}
+ 
+.header {
+    background-image: url(/apple-pie/images/apple-pie.jpg);
```
joaopdg, line 12: To get your background image to look a bit more like the lab example you can do this:

```
.header {
    background-image: url(/apple-pie/images/apple-pie.jpg);
    background-position: 0 0;
    background-size: cover;
    background-repeat: no-repeat;
    height: 450px;
    width: 100vw;
    padding-top: 50px;
    position: relative;
  }
```

Delete some lines of code that aren't doing anything and change the `width` to `100vw` this way the image will have the same with as the screen.

apple-pie/styles/style.css
```diff
@@ -0,0 +1,68 @@
+
+* {
+	box-sizing: border-box;
+	margin: 0 10%;
```
joaopdg, line 4: To get rid of the white bars around the background image change your margin here to this `margin: 0px;`

After this you can adjust your left margins on the text to center them a little bit more.

joaopdg (commented):
Great job on the lab Eveline!

PR #3536, opened 2023-01-18
[LISFT0123] Pedro Leal

PeterLoyal:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,127 @@
...
+  font-size: 30px;
+  margin-top: 200px;
+  margin-bottom: 200px;
+  line-height: 50px;
+  text-align: center;
+}
+
+.imgRecipe{
+  width: 600px;
+  margin-left: 350px;
```
webLisbonTA, line 45: Giving margin only on one side it's not a good pratice to center the image.

apple-pie/styles/style.css
```diff
@@ -0,0 +1,127 @@
...
+
+.secondSec{
+  
+  display: flex;
+	flex-direction: column;
+	flex-wrap: nowrap;
+	justify-content: center;
+	align-items: center;
+	align-content: center;
+  margin-right: 650px;
```
webLisbonTA, line 88: Here you should've ajusted the margin with the side that it is closer.

webLisbonTA (commented):
Good job on the lab. Give a look at the comments I left you.

PR #3541, opened 2023-01-18
 [LISFT0123]Farid 

Farid-97:


lzaquine (commented):
Good job on the lab Farid. One piece of advice, try giving the elements IDs or classes so it's organized and you can target them on CSS!

PR #3542, opened 2023-01-19
[LISFT0123] Caio Mazur

caiomazur:


apple-pie/index.html
```diff
@@ -4,18 +4,69 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Apple Pie Recipe</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="styles/style.css">
   </head>
   <body>
-    Apple Pies This was my grandmother's apple pie recipe. I have never seen another one quite like it. It will always
+
+    <section id="title-section">
```
joaopdg, line 11: Here you could have used the `<header> </header>` tag instead of `<section> </section>` the result will be the same, it's just a matter of semantics and good practice, and it is important for people that use screen readers.

joaopdg (commented):
Great job on the lab Caio!

PR #3544, opened 2023-01-19
[LISFT0123] Rafaela Umbelino

RafaelaUmbelino:


apple-pie/styles/style.css
```diff
@@ -0,0 +1,82 @@
...
+  align-items: center;
+  margin-bottom: 50px;
+  
+  
+}
+
+ul > li {
+  list-style: none;
+  margin: 5px;
+  margin-left: 550px;
```
webLisbonTA, line 55: You used too much margin-left, so the text is at the right side of the screen.

webLisbonTA (commented):
Good job on the lab.

PR #3646, opened 2023-03-20
Gerrit Brinsat FT MAR 23 BER

GBrinsat:
Add completed HTML and CSS


GBrinsat:
Add French Soup lab practice

PR #3730, opened 2023-05-08
[FTRMT052023] cédricRocher

hymced:


hymced:
> how to exactly align ol and ul ? if I used a left margin of 20px for ol as for ul, but there is an offset, so I used 15px for ol because it visually seems to be a good value, but how to be sure?

solved, the markers of the ul has to be inside the li elements (as the counted markers of the ol) instead of outside

PR #3753, opened 2023-05-08
modified index.html & style.css

linhvnde:
im not sure if the font is working correctly as the font given doesnt look similar to the demo picture


TA-Remote:
hey Linh, great job!!!🙌
same font can look differently in a different browser, that's why they use to add some alternatives that look similar. So if the first doesn’t work you have another one that almost looks the same (otherwise the browser will use the default font). 
 —> font-family: "Helvetica" (1st) , sans-serif (option);
 anyway it looks fine to me!!!😉
 Alessandra

PR #3874, opened 2023-09-22
done

VicJunquera:


VicJunquera:
Now with the correct images' paths in index.html

PR #3886, opened 2023-10-09
[LIS-FT-OCT23] Guilherme Silva

51V3:


webLisbonTA:
Hi Guilherme, 
Great job on the lab!

Please note that you have the ingredients ordered, and on that part they were supposed to be with "-".

One way to do it would be:

#nameOfTheId li:before {
  content: "- ";
  line-height: 1.3;
}

Overall, good job!

PR #3888, opened 2023-10-09
[LIS-FT-OCT23] caroline_vallim

CarolineVallim:


webLisbonTA:
Hi Carol,
Great job on the lab!
The first image is a little small, in the h1 you could do a little more padding, padding: 250px - for example.

Overall, great job!

PR #3892, opened 2023-10-10
[LIS-FT-OCT23] filipe_dias

filipedias2000:


webLisbonTA:
Hi Filipe,
Great job on your lab!

Great job on this lab!
Another way of solving the "-" on the ingredients, would be through CSS.
You would have to create a class on the parent of "li", and then:

."parent" li:before (This is an example, it would depend on what you have on the css code) {
content: "- ";
}

Overall, great job!

PR #3893, opened 2023-10-10
[LIS-FT-OCT23] David_Lucca

davidvlucca:


webLisbonTA:
Hi David,
Great job on your lab!

Please note that you do not have to give a class to your h1, for example, you could use the following syntax:
.section-header h1 - With this syntax you are saying that you want to access the h1 on your section-header class.

You are naming all the classes and you can avoid that with this syntax.

Overall, great job!

PR #3894, opened 2023-10-10
[LIS-FT-OCT23] daniel_rodrigues

daanrxz:


webLisbonTA:
Hi Daniel,
Great job on the lab!

Some things to improve:
Your intro-text class is centered. If you use the below would be better:
.container {
  width: 60%;
  margin-left: auto;
  margin-right: auto;
}

The directions text is on the right side of your page and that is because you have too much on the margin-left.

Overall, great job!

PR #3895, opened 2023-10-10
[LIS-FT-OCT23] Bruno_M

brrrun:


webLisbonTA:
Hey Bruno,
Great job on this lab! 
Another way of solving the "-" on the ingredients, would be through CSS:

.boxes li:before (This is an example, it would depend on what you have on the css code) {
  content: "- ";
}

PR #3899, opened 2023-10-10
 [LIS-FT-OCT23] miguel_mestre

trombinhas:


jacobviv (commented):
Good work with your lab Miguel

apple-pie/styles/style.css
```diff
@@ -21,4 +21,56 @@ body {
...
 /* STYLES */
-/* Write your CSS below */
\ No newline at end of file
+/* Write your CSS below */
+body{
+    margin-left: 60px;
+}
+
+.header{
+  background-image: url(apple-pie.jpg);
```
jacobviv, line 30: Images should be in the images folder, then you could route the css like this
```css
background-image: url("./../images/apple-pie.jpg");
```

apple-pie/styles/style.css
```diff
@@ -21,4 +21,56 @@ body {
 }
 
 /* STYLES */
-/* Write your CSS below */
\ No newline at end of file
+/* Write your CSS below */
+body{
+    margin-left: 60px;
```
jacobviv, line 26: This margin property in the body is why your header image doesn't have the screen width

PR #3939, opened 2023-10-17
FTRMT102023 Elise Jonkers

elisejonkers:
I had some issues with the following and I would like to have specific feedback on those topics:
- Creating the right size of the background image;
- Creating the right sizes using the box model.


TA-Remote:
hello @elisejonkers , 
nice job with your solution🙌
to give to the background-img the size of an block element you can use the css property width: 100%. the element will take the full width of its parent element. (keep in mind default margin and paddind --> to check you can always use the dev tools in the browser), then you can apply margin to set the element size. 
Hope you can find this tip useful... anyway there are a lot of different css properties you can play with in order to have the same behaviour😉

keep it up!!!
Alessandra

PR #3947, opened 2023-10-21
Final apple pie changes

BanashreeAribenchi:
- The background image of the Hero section is center-aligned.
- The font type and font size are changed according to the requirements.


BanashreeAribenchi:
done

